Issue #67, opened 2014-08-09
php-vcr unit test fail on Windows 7


b01:
Trying to run unit test on latest master branch fails. Could it be that PHP-VCR is overwriting its own class?

On windows 7, with PHP 5.6 RC3

``` bash
Khalifah@KHALIFAH-PC /e/Khalifah/Projects/php-vcr (master)
$ phpunit
PHPUnit 3.7.37 by Sebastian Bergmann.

Configuration read from E:\Khalifah\Projects\php-vcr\phpunit.xml

...............................................................  63 / 198 ( 31%)
...............
Fatal error: Class 'VCR\Util\SoapClient' not found in E:\Khalifah\Projects\php-vcr\src\VCR\Util\SoapClient.php on line 12

Call Stack:
    0.0000     124248   1. {main}() C:\Users\Khalifah\vendor\phpunit\phpunit\phpunit:0
    0.0040     326056   2. PHPUnit_TextUI_Command::main() C:\Users\Khalifah\vendor\phpunit\phpunit\phpunit:54
    0.0040     326384   3. PHPUnit_TextUI_Command->run() C:\Users\Khalifah\vendor\phpunit\phpunit\src\TextUI\Command.php:138
    0.7470    3746792   4. PHPUnit_TextUI_TestRunner->doRun() C:\Users\Khalifah\vendor\phpunit\phpunit\src\TextUI\Command.php:186
    0.8020    3926400   5. PHPUnit_Framework_TestSuite->run() E:\Khalifah\Projects\php-vcr\vendor\phpunit\phpunit\PHPUnit\TextUI\TestRunner.php:350
    1.7681    6218960   6. PHPUnit_Framework_TestSuite->run() E:\Khalifah\Projects\php-vcr\vendor\phpunit\phpunit\PHPUnit\Framework\TestSuite.php:709
    1.7681    6219440   7. PHPUnit_Framework_TestSuite->runTest() E:\Khalifah\Projects\php-vcr\vendor\phpunit\phpunit\PHPUnit\Framework\TestSuite.php:
749
    1.7681    6219792   8. PHPUnit_Framework_TestCase->run() E:\Khalifah\Projects\php-vcr\vendor\phpunit\phpunit\PHPUnit\Framework\TestSuite.php:779
    1.7691    6220040   9. PHPUnit_Framework_TestResult->run() E:\Khalifah\Projects\php-vcr\vendor\phpunit\phpunit\PHPUnit\Framework\TestCase.php:783
[...]

PR #70, opened 2014-08-23
php-vcr issue #67: unit test fail on Windows 7.

b01:


phpunit.xml
```diff
@@ -1,5 +1,4 @@
 <phpunit bootstrap="tests/bootstrap.php"
```
dbu: the phpunit.xml file should be renamed to phpunit.xml.dist and phpunit.xml be in .gitignore. then its much easier to customize the phpunit file.

phpunit.xml
```diff
@@ -1,5 +1,4 @@
...
-         colors="true"
```
dbu: that sounds like a phpunit bug - if it can't output colors it should ignore this. what version of phpunit do you have?

renatomefidf:
@b01 sorry for the delay, sadly due to that your PR got outdated. Are you still willing to work on this?
If yes maybe you could keep the colors true on the phpunit config. I'll check better the PR if you want to proceed!

Thank you for your time and sorry again for the delay.

renatomefi:
Since no one else complained about it and the author is not responding I'm closing this for now.
If anyone wants to make it again please fell free to open a new PR!

Thank you all

PR #86, opened 2014-12-20
vcr hooks implementation

aaa2000:
The implementation of new features is more easier with these hooks.
https://relishapp.com/vcr/vcr/v/2-9-3/docs/hooks

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
     "require": {
         "ext-curl": "*",
         "beberlei/assert": "2.*",
-        "symfony/yaml": "2.*"
+        "symfony/yaml": "2.*",
+        "symfony/event-dispatcher": "~2.1"
     },
 
     "require-dev": {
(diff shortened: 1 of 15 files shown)


K-Phoen:
Hell yeah!
It would simplify the bundle's code and even allow a few nice things like telling the user which requests were replayed and which ones were not.

I didn't review @aaa2000's code but I'm definitely :+1: for events.

tylercollier:
@adri You asked if YAGNI. The Ruby VCR version has this, why wouldn't we want it here? Hooks are perfect for the "not everyone will need it, but some will" case.

adri:
@aaa2000 Could you please rebase on master?

aaa2000:
Done

adri:
Merged. Great work, thank you @aaa2000.

aaa2000:
Thanks, but I had the symfony/event-dispatcher dependency because Guzzle use it. But now, the dependency with Guzzle was removed, maybe later you should reimplement a VCR event dispatcher.

adri:
Because you worried about dependency conflicts with other projects? 
I suspect the EventDispatcher to be pretty stable by now.

K-Phoen:
Bam, the VCRBundle has been updated to use the new hooks!

aaa2000:
yes, dependency conflicts with other projects when a version 3 of EventDispatcher will be publish.

Issue #85, opened 2014-12-16
Wrong JSON stored in cassette


alexdomsa:
I have a test which uses file "testSearch.json" as cassette.

If the file does not already exist before running the test, the JSON stored in it is correct:
```
[{ .... }]
```

If the file already exists and is blank, will store a wrong JSON:
```
{ ... }]
```
We can see that the leading square bracket is missing

adri:
Fixed with #91.

PR #91, opened 2015-01-11
Correction of JSON when the file already exists and is blank

aaa2000:
It is a PR for #85.


adri:
Looks good to me, thank you!

PR #93, opened 2015-01-15
make single files for unique requests

chregu:
This makes the whole thing a lot faster if you have huge files

Not sure, if this is the proper way, but take it as a proof of concept. I'm doing a PR just that it doesn't get forgotten.

### DIFF
--- a/src/VCR/Videorecorder.php
+++ b/src/VCR/Videorecorder.php
@@ -171,9 +171,14 @@ class Videorecorder
                 . "VCR::insertCassette('name');"
             );
         }
+        $cassettename = $this->cassette->getName();
+        $cassettename = pathinfo($cassettename, PATHINFO_FILENAME);
+        $md5 = md5(md5($request->getBody()).md5($request->getUrl()).$request->getMethod());
+        $this->insertCassette($cassettename . "." . $md5);
 
-        if ($this->cassette->hasResponse($request)) {
-            return $this->cassette->playback($request);
+        $response = $this->cassette->playback($request);
+        if ($response !== null) {
+             return $response;
         }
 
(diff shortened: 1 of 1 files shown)


adri:
Thanks! Ideally each test should choose its own cassette file which is named after whatever the test is supposed to test. 

For example using PHPUnit annotations `@vcr endpoint-xy-fails-with-error.json`. Would collect all requests for this exact test case. Of course there could still be many requests issued in one test. For performance optimization I'm thinking about keeping all requests in memory as long as they fit within the `memory_limit`.

tylercollier:
I like the idea.

I know it was a proof of concept. I think I'd store the hash in each request in the cassette, alongside the method/header/body/etc. The way it is currently would mean that the filename itself would change with each content change. You'd have to go change the tests/annotations to point to the right file. This would also be annoying (to me) to see in a the logs of git.

This idea would make it harder to change a cassette by hand. I'm not sure if anyone else does it manually, but it's typically how I maintain my cassettes (small updates to url or body). Additionally, my cassettes rarely have more than 3 requests in them and they are small payloads, so forcing a hash would be annoying. I recommend it be implemented as a feature: if a stored hash is present, use it to verify a match, but otherwise, verify a match the way it's been done (by verifying the headers/method/body/etc separately).

dbu:
we determine our name dynamically (in a symfony project): 

```
            $fixtureName = $request->cookies->get('X-BehatTestName',
                $this->getFixtureName($request->server->get('REQUEST_URI'))
            );
```

for this, we first load favicon.ico to get phantomjs initialized on the site, then set that cookie to tell the fixture name and start the actual test.

adri:
I'm going to close this :-) Dynamically setting the cassette name from outside is possible (btw. now with folder support in cassette names, #107) and general speed improvements should be solved on a storage level.

PR #97, opened 2015-01-28
Various fixes

jamiehannaford:
This PR introduces two modifications/fixes:

1. `Request::getHost` does not include the port, which means that request matching fails for various cassettes that contain hostnames with an explicit port. See the [RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23)

2. `RequestMatcher::matchHeaders` fails to match requests which have the same null headers. But IMO if two requests have `Accept: null` that denotes equality. This is also causing cassette matching problems

### DIFF
--- a/src/VCR/Request.php
+++ b/src/VCR/Request.php
@@ -197,7 +197,13 @@ class Request
      */
     public function getHost()
     {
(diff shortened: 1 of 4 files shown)


adri:
Very nice Jamie, thank you for you contributions!

PR #99, opened 2015-02-03
Better request header matching

tylercollier:
`isset()` is not a good choice if the header is null. Headers shouldn't be null, but Guzzle is giving me null headers in my application. However, then `isset($firstHeaders[$key])` is false, and thus it returns false. I've corrected this with `array_key_exists`.

Second, the checks for headers should be reciprocal. We should be checking that headers on the current request aren't missing from the cassette's request. My code isn't perfectly efficient... you could keep track of which comparisons have already been done. But it it works and is clear to read and understand I think.

---

On the `url_stat` stuff, it's a separate issue enitrely. I'm not sure how to exclude it from the pull request. I asked a question on how to deal with it [on stackoverflow](http://stackoverflow.com/questions/25296178/why-would-phps-stat-throw-a-runtimeexception).


tylercollier:
Moved to https://github.com/php-vcr/php-vcr/pull/100

Issue #33, opened 2014-01-20
StreamProcessor: Stat missing file causes exception


kasperg:
I have a project which [autoloads classes for a namespace from multiple directories](https://github.com/wsdl2phpgenerator/wsdl2phpgenerator/blob/master/composer.json#L10).

After integrating PHP-VCR [an exception is thrown when the Composer autoloader first tries to include the class](https://travis-ci.org/wsdl2phpgenerator/wsdl2phpgenerator/jobs/17256236) from the first directory (where the file is not located) before trying in the second directory. This causes my tests to fail.

As far as I can tell the problem seems to be related to [`StreamProcessor->url_stat()`](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Util/StreamProcessor.php#L190) but here errors caused by the `stat` call should be suppressed.

Notes:
- I have only seen this when running tests on Travis across PHP versions 5.3.3-5.5 - not on my local machine (PHP 5.3.27)
- [This behavior does not occur when PHP-VCR is disabled](https://travis-ci.org/wsdl2phpgenerator/wsdl2phpgenerator/jobs/17261083).

Any idea what could cause this?

PR #100, opened 2015-02-05
Better header matching

tylercollier:
**Update from previous pull request: Used branch, added test.**

`isset()` is not a good choice if the header is null. Headers shouldn't be null, but Guzzle is giving me null headers in my application. However, then `isset($firstHeaders[$key])` is false, and thus it returns false. I've corrected this with `array_key_exists`.

Second, the checks for headers should be reciprocal. We should be checking that headers on the current request aren't missing from the cassette's request. My code isn't perfectly efficient... you could keep track of which comparisons have already been done. But it it works and is clear to read and understand I think.

---

On the `url_stat` stuff, it's a separate issue enitrely. I'm not sure how to exclude it from the pull request. I asked a question on how to deal with it [on stackoverflow](http://stackoverflow.com/questions/25296178/why-would-phps-stat-throw-a-runtimeexception), so you can ignore it here. I'm sorry I'm new to pull requests like this, but I found some [documentation on how to rebase a pull request to ignore this part](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request).


adri:
Hey Tyler, thank you very much for this. I've simplified the request matching in [this commit]( 394737a317f79abe1c7bdbe15621fe45700e24b7) and took over your tests. 

What is the the `url_stat` stuff about? :-) Can you create another issue for it please? It might be related to #33.

adri:
Released [1.2.2](https://github.com/php-vcr/php-vcr/releases/tag/1.2.2)

Issue #96, opened 2015-01-26
Does PHP-VCR work with stream_context_create?


tylercollier:
I'm using the [expert-sender-api](https://github.com/LinguaLeo/expert-sender-api) library. The request in the cassette that ends up recorded is almost blank:

```
    request:
        method: GET
        url: [URL HERE]
        headers:
            Host: [HOST HERE]
            Content-Length: '0'
```

There should be a request body, and the content-length should not be zero (it's also odd it's a string, no?). Also, the method there is GET, but it should be POST. If I disable PHP-VCR, and perform a live web service call, the request goes out as I expect and then I get a proper response. So something is not right or fully handled by PHP-VCR here.

If I enabled PHP-VCR and let it send the blank request, it captures the response. As you might expect, it's a response to the blank request, but in my case, the response contains 9 headers and an XML body that I'd expect. So that part works.

The expert-sender-api library's `query()` method [here](https://github.com/LinguaLeo/expert-sender-api/blob/master/src/LinguaLeo/ExpertSender/HttpTransport.php#L12) makes use of PHP's `stream_context_create` I'm wondering if that has anything to do with it. I suppose not, because that's part of HTTP wrapper, which [PHP-VCR explicitly states it supports](http://php-vcr.github.io/documentation/libraryhooks/).

If I manually create the cassette request, there's still another problem. Weirdly (this is directly from the [php manual](http://php.net/manual/en/reserved.variables.httpresponseheader.php)):

> When using the HTTP wrapper, $http_response_header will be populated with the HTTP response headers. $http_response_header will be created in the local scope.

The fatal error I get with PHP-VCR is:

```
Undefined variable: http_response_header
```

That variable is from [line 32](https://github.com/LinguaLeo/expert-sender-api/blob/master/src/LinguaLeo/ExpertSender/HttpTransport.php#L32). Is it possible for PHP-VCR to handle this? I know essentially how PHP-VCR works, and I can't imagine it dealing with that. Searching the source code for `http_response_header` yields nothing.

tylercollier:
Regarding the "there's still another problem". I just ran a different test with PHP-VCR off and not cassette loaded. I received a response using `file_get_contents`, but PHPUnit still choked, saying `Undefined variable: http_response_header`. I'm looking more into this mysterious `http_response_header` thing in PHP, and I'll report back.

adri:
[...]

PR #105, opened 2015-02-23
Adds stream wrapper support for intercepting http_response_header local variable

tylercollier:
Motivation for this is #96. Adrian suggested using the code transformer to deal with this. I did. The only concern I had was if the stream wrapper runs synchronously such that we're able to store off the most recent response during playback, and then access it when intercepting `$http_response_header`. I wrote a test to show that too.

To do the transform, I tried to follow the pattern from the existing curl and soap transformers, which take  `AbstractCodeTransform` and `StreamProcessor` arguments to their constructors in their `LibraryHook` implementations. However, since `StreamWrapperHook` is instantiated by PHP internals (well, we do it first, and then later PHP does), PHP doesn't pass any values to the constructor. I got around this as you can see, but think it'd be better to pass the `AbstractCodeTransform` and `StreamProcessor` to a setter from the `createVCRLibraryHooksStreamWrapperHook` factory method, and then call another function on `StreamWrapperHook` to invoke register/intercept. Thoughts?

Notes:
* Be sure to update the composer autoload. To do this, run `php composer.phar dump-autoload`.
* I couldn't get the integration tests to run. When I run all tests with `phpunit ./tests`, it says

        PHP Fatal error:  Class 'VCR\Example\ExampleHttpClient' not found in /home/tylercollier/repos/php-vcr/tests/integration/guzzle/test/VCR/Example/ExampleHttpClientTest.php on line 50

    I saw the extra README file in that directory but couldn't make sense of what it might be trying to tell me to do.
* I removed the `unittest_guzzle_test` cassette, since it wasn't being used. Actually I renamed it to `unittest_post_test` and now I am using it.
* I had to remove the `Content-Length` header in the `unittest_curl_test` cassette or else the `testShouldInterceptCurlLibrary` test wouldn't pass. Is that due to differing PHP curl library versions or something? I have 7.22.0. In that test, it doesn't set that header directly, so where does it come from?

### DIFF
--- a/src/VCR/CodeTransform/StreamProcessorCodeTransform.php
+++ b/src/VCR/CodeTransform/StreamProcessorCodeTransform.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace VCR\CodeTransform;
+
+class StreamProcessorCodeTransform extends AbstractCodeTransform
+{
+    const NAME = 'vcr_streamprocessor';
+
+    private static $patterns = array(
+        // Substitute for the $http_response_header variable, unless it's used as an assignment.
+        '/\$http_response_header\b(?!\s*=)/i' => '\VCR\LibraryHooks\StreamWrapperHook::getLastResponseHeaders()'
+    );
+
+    /**
+     * @inheritdoc
+     */
+    protected function transformCode($code)
+    {
+        return preg_replace(array_keys(self::$patterns), array_values(self::$patterns), $code);
+    }
+}
--- a/src/VCR/LibraryHooks/StreamWrapperHook.php
+++ b/src/VCR/LibraryHooks/StreamWrapperHook.php
@@ -4,9 +4,11 @@ namespace VCR\LibraryHooks;
 
 use VCR\Request;
 use VCR\Response;
+use VCR\CodeTransform\AbstractCodeTransform;
 use VCR\Util\Assertion;
 use VCR\Util\HttpUtil;
 use VCR\Util\StreamHelper;
+use VCR\Util\StreamProcessor;
 
 /**
  * Library hook for streamWrapper functions using stream_wrapper_register().
@@ -39,10 +41,45 @@ class StreamWrapperHook implements LibraryHook
     public $context;
 
     /**
+     * @var AbstractCodeTransform
+     */
+    private static $codeTransformer;
+
+    /**
+     * @var StreamProcessor
+     */
+    private static $processor;
+
+    /**
+     * @var Response Last response processed.
+     */
+    private static $lastResponse;
+
+    /**
+     * Creates a new Stream Wrapper hook instance.
+     *
+     * @param AbstractCodeTransform  $codeTransformer
+     * @param StreamProcessor $processor
+     */
+    public function __construct($codeTransformer = null, $processor = null)
+    {
+        if ($codeTransformer) {
+            self::$codeTransformer = $codeTransformer;
+        }
+        if ($processor) {
+            self::$processor = $processor;
+        }
+    }
+
+    /**
      * @inheritDoc
      */
     public function enable(\Closure $requestCallback)
     {
+        self::$codeTransformer->register();
+        self::$processor->appendCodeTransformer(self::$codeTransformer);
+        self::$processor->intercept();
+
         Assertion::isCallable($requestCallback, 'No valid callback for handling requests defined.');
         self::$requestCallback = $requestCallback;
         stream_wrapper_unregister('http');
@@ -91,6 +128,7 @@ class StreamWrapperHook implements LibraryHook
 
         $requestCallback = self::$requestCallback;
         $this->response = $requestCallback($request);
+        self::$lastResponse = $this->response;
 
         return true;
     }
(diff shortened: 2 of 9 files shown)


adri:
I'm sorry but maybe we misunderstood each other. I was just brainstorming how it can be done and a reason for not replacing the `$http_response_header` in a code transformer was the false positives with other people using this variable in their code. How would you get around that? 

Yeah the integration tests are separate composer projects because they install external dependencies (guzzle). I see this needs more documentation :-)

tylercollier:
I reasoned that we can ignore transforming the code where `$http_response_header` is being assigned to. But I suppose this doesn't help, since people will still need to read from that variable.

What about this? Do you see how the transformation reads from `StreamWrapperHook::getLastResponseHeaders()`? We would clear out the last response headers for each test. Then we would change the transform to be something like `(StreamWrapperHook::getLastResponseHeaders() ?: [original code here])`. I haven't had time to test if this is feasible yet, but I will.

Your other idea of using PHP-Parser is a good one. A simpler approach is better if it works though, so I'll look into the above.

tylercollier:
I reasoned that we can ignore transforming the code where `$http_response_header` is being assigned to. But I suppose this doesn't help, since people will still need to read from that variable.

What about this? Do you see how the transformation reads from `StreamWrapperHook::getLastResponseHeaders()`? We would clear out the last response headers for each test. Then we would change the transform to be something like `(StreamWrapperHook::getLastResponseHeaders() ?: [original code here])`. I haven't had time to test if this is feasible yet, but I will.

Your other idea of using PHP-Parser is a good one. A simpler approach is better if it works though, so I'll look into the above.

adri:
What would happen if someone uses `$http_response_header` outside of the local scope but still within the same test? 

I agree, it would be great to have a simple solution to it. It's a though problem, to be honest I don't know (yet) how to make it work with a simpler approach without introducing bugs. I keep thinking about it. 
Maybe its possible to set another variable in the local scope when a request is made and then indeed make a check like you proposed and see if that variable is there. A bit hacky though ;-) There it also depends if we can even set a variable properly.

tylercollier:
I think the answer is to set `$http_response_header` ourselves. I haven't tried this yet, but in a code transform, we'd look for e.g. `file_get_contents` and set the response header in a line after that. The best part of this solution I think is that it mimics how PHP does it (compared to using a code transform that changes the READ of that local variable, which I what I was doing before).

That would handle your concern of the use of `$http_response_header` outside of the local scope but still within the same test:

```
function doAnotherNetworkThing() {
    $result = @file_get_contents($url, false, $context);
    $http_response_header = getLastResponseHeader(); // This line created by code transform
    $q = $http_response_header;
    // $q is ['c' => 'd']
}

$result = @file_get_contents($url, false, $context);
$http_response_header = getLastResponseHeader(); // This line created by code transform
$x = $http_response_header;
// $x is ['a' => 'b']
$this->doAnotherNetworkThing()
$y = $http_response_header;
// $y is ['a' => 'b'], good!
```

And it would handle sequential local scope network access:

```
$result = @file_get_contents($url, false, $context);
$http_response_header = getLastResponseHeader(); // This line created by code transform
$x = $http_response_header;
// $x is ['a' => 'b']
$result = @file_get_contents($url, false, $context);
$http_response_header = getLastResponseHeader(); // This line created by code transform
$y = $http_response_header;
// $y is ['c' => 'd'], good!
```

Thoughts?

To address the concern that this would be buggy. We could make this opt-in, by requiring that users set a configuration value saying they know what they're doing. As long as someone is aware of what's going on, they'd be able to help us track down bugs. It would be great if we could detect situations where our fix can't be guaranteed to work and warn the user (or perhaps throw).

The only bug I can think of for the above proposed solution is if file_get_contents is used for a local file, and they happen to set http_response_headers themselves. In that case, we could transform the call to file_get_contents to proxy through a function of ours. In that function, we could keep track of whether the url supplied was local or not, and change getLastResponseHeader's response appropriately.

tylercollier:
** The following are notes for myself **

> What would happen if someone uses $http_response_header outside of the local scope but still within the same test?

Good point. Here's an example where that would be a problem:

```
function doAnotherNetworkThing() {
    // pretend network action happened on this line
    $q = $http_response_header;
    // $q is ['c' => 'd']
}

// pretend network action happened on this line
$x = $http_response_header;
// $x is ['a' => 'b']
$this->doAnotherNetworkThing()
$y = $http_response_header;
// $y is ['c' => 'd'] but should be ['a' => 'b']
```

I found a way around that (not good enough, I explain below):

```
function doAnotherNetworkThing() {
    // pretend network action happened on this line
    $q = isset($php_vcr_http_response_header) ? $php_vcr_http_response_header : $php_vcr_http_response_header = getLastResponseHeader();
    // $q is ['c' => 'd']
}

// pretend network action happened on this line
$x = isset($php_vcr_http_response_header) ? $php_vcr_http_response_header : $php_vcr_http_response_header = getLastResponseHeader();
// $x is ['a' => 'b']
$this->doAnotherNetworkThing()
$y = isset($php_vcr_http_response_header) ? $php_vcr_http_response_header : $php_vcr_http_response_header = getLastResponseHeader();
// $y is ['a' => 'b'], good!
```

But that breaks down for another local scope network access:

```
// pretend network action happened on this line
$x = isset($php_vcr_http_response_header) ? $php_vcr_http_response_header : $php_vcr_http_response_header = getLastResponseHeader();
// $x is ['a' => 'b']
// pretend network action happened on this line
$y = isset($php_vcr_http_response_header) ? $php_vcr_http_response_header : $php_vcr_http_response_header = getLastResponseHeader();
// $y is ['a' => 'b'], should be e.g. ['c' => 'd']
```

PR #106, opened 2015-02-23
Add flags parameter for url_stat call on stream wrapper

dbu:
according to http://php.net/manual/en/streamwrapper.url-stat.php url_stat has a second $flags parameter. if that is not passed through, checks like is_link fail

### DIFF
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -258,7 +258,11 @@ class StreamProcessor
      */
     public function stream_stat()
     {
-        return fstat($this->resource);
+        $this->restore();
+        $stat = fstat($this->resource);
+        $this->intercept();
+
+        return $stat;
     }
 
     /**
@@ -280,15 +284,16 @@ class StreamProcessor
      *
      * @link http://www.php.net/manual/en/streamwrapper.url-stat.php
      *
-     * @param  string $path The file path or URL to stat.
+     * @param string $path  The file path or URL to stat.
+     * @param int    $flags Holds additional flags set by the streams API.
      *
-     * @return integer      Should return as many elements as stat() does.
+     * @return integer Should return as many elements as stat() does.
      */
-    public function url_stat($path)
+    public function url_stat($path, $flags)
     {
         $this->restore();
         try {
-            $result = @stat($path);
+            $result = @stat($path, $flags);
         } catch (\ErrorException $e) {
             // PHPUnit running in process isolation (processIsolation="true")
             // throws an \ErrorException for any PHP warning.


src/VCR/Util/StreamProcessor.php
```diff
@@ -258,7 +258,11 @@ public function stream_seek($offset, $whence = SEEK_SET)
      */
     public function stream_stat()
     {
-        return fstat($this->resource);
+        $this->restore();
+        $stat = fstat($this->resource);
+        $this->intercept();
```
dbu: any idea why this messes up the tests as much as it does?

anyways, not very sure about this change. but without it i got weird errors too

dbu:
ok so this is all wrong. `stat` has no second parameter. no idea if it should matter or not at all. but it seems that with the `@` we just do not see that everything stops working.

dbu:
the whole problem turned up when trying to use the vcr-bundle in symfony, which loads the recorder even for the console. this playes hell with composer installing things and the cache, starting to not see files that are there, getting confused by symlinks and other annoying problems.

maybe we indeed need to look at the $flags and do more than just `stat`.

i think we will just go back to load VCR in our kernel during the `handle` when in test env again, instead of using the bundle. that way we avoid the cli and things seem to more or less work.

src/VCR/Util/StreamProcessor.php
```diff
@@ -280,15 +284,16 @@ public function stream_tell()
...
-     * @return integer      Should return as many elements as stat() does.
+     * @return integer Should return as many elements as stat() does.
      */
-    public function url_stat($path)
+    public function url_stat($path, $flags)
     {
         $this->restore();
         try {
-            $result = @stat($path);
+            $result = @stat($path, $flags);
```
dbu: `stat` has no second argument, so we need to do that work ourselves it seems. no idea what we really have to do.

dbu:
see #108

adri:
The vcr-bundle is only intended to show the HTTP requests an app makes in the web profiler. It doesn't store any fixtures.

The main issue is indeed #108 or something similar related to the stat/url_stat code.

@K-Phoen Do you think it makes sense to not load VCR on the console?

dbu:
well, the bundle does [start the videorecorder](https://github.com/php-vcr/VCRBundle/blob/master/Resources/config/services.yml#L31-L32) and this is explicitly triggered in the bundle boot method.

and in general, VCR can make sense on the console. assume i want to functional test a service that does a web request in phpunit for example.

i think in our project the other stumble block is missing flexibility how to determine the cassette name. we will just instantiate the VCR in the AppKernel when the environment is "test".

adri:
When you *test* something like a command, the vcr-bundle is not necessarily the right choice. Maybe you can trick it into using your own configuration but primarily it is used while *developing* to see request the app does. It integrates into the web profiler to show all requests. @K-Phoen please correct me if I'm wrong.

To me these are different use cases.

The way you go now with initiating VCR in the AppKernel for env test is in my opinion the right way if you do (Behat) tests with selenium where the app and the test code run in different processes.

dbu:
i can totally live with it.

but it would be convenient if the bundle could work that way so i don't
need to touch my kernel. so i don't think it would be wrong to make this
more useful, but totally not a priority or anything.

and btw, i occasionally look at profiler recordings from a cli run when
i want to debug what the command was doing.

K-Phoen:
@adri the bundle have two use cases:
* showing the requests made by an app in the profiler (would be a display-only mode),
* record the requests made by the app to replay them later in tests.

That's why I provided a way to configure the storage used for the cassettes (blackhole would be for the display-only mode for instance).

PR #107, opened 2015-02-23
support path separators in cassette names

dbu:
This change makes it possible to group fixtures by folder names.

To work correctly on windows systems, you need to use PATH_SEPARATOR rather than hardcode `/`.

### DIFF
--- a/src/VCR/Storage/AbstractStorage.php
+++ b/src/VCR/Storage/AbstractStorage.php
@@ -3,6 +3,7 @@
 namespace VCR\Storage;
 
 use VCR\Util\Assertion;
+use VCR\VCRException;
 
 /**
  * Abstract base for reading and storing records.
@@ -50,12 +51,21 @@ abstract class AbstractStorage implements Storage
     /**
      * Creates a new file store.
      *
+     * If the cassetteName contains PATH_SEPARATORs, subfolders of the
+     * cassettePath are autocreated when not existing.
+     *
      * @param string $cassettePath Path to the cassette directory.
-     * @param string $cassetteName Path to a file, will be created if not existing.
+     * @param string $cassetteName Path to the cassette file, relative to the path.
      */
     public function __construct($cassettePath, $cassetteName, $defaultContent = '[]')
     {
-        $file = $cassettePath . DIRECTORY_SEPARATOR . $cassetteName;
+        Assertion::directory($cassettePath, 'Cassette path "{$cassettePath}" is not existing or not a directory');
+
+        $file = rtrim($cassettePath, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $cassetteName;
+
+        if (!is_dir(dirname($file))) {
+            mkdir(dirname($file), 0777, true);
+        }
 
         if (!file_exists($file) || 0 === filesize($file)) {
             file_put_contents($file, $defaultContent);
@@ -104,6 +114,8 @@ abstract class AbstractStorage implements Storage
      */
     public function __destruct()
     {
-        fclose($this->handle);
+        if (is_resource($this->handle)) {
+            fclose($this->handle);
+        }
     }
 }
--- a/tests/VCR/Storage/AbstractStorageTest.php
+++ b/tests/VCR/Storage/AbstractStorageTest.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace VCR\Storage;
+
+use org\bovigo\vfs\vfsStream;
+
+/**
+ * Test integration of PHPVCR with PHPUnit.
+ */
+class AbstractStorageTest extends \PHPUnit_Framework_TestCase
+{
+    protected $handle;
+    protected $filePath;
+    protected $storage;
+
+    public function testFilePathCreated()
+    {
+        $fs = vfsStream::setup('test');
+
+        $this->storage = new TestStorage(vfsStream::url('test/'), 'file');
+        $this->assertTrue($fs->hasChild('file'));
+
+        $this->storage = new TestStorage(vfsStream::url('test/'), 'folder/file');
+        $this->assertTrue($fs->hasChild('folder'));
+        $this->assertTrue($fs->getChild('folder')->hasChild('file'));
+    }
+
+    /**
+     * @expectedException \VCR\VCRException
+     */
+    public function testRootNotExisting()
+    {
+        vfsStream::setup('test');
+        $this->storage = new TestStorage(vfsStream::url('test/foo'), 'file');
+    }
+
+}
+
+class TestStorage extends AbstractStorage
+{
+    private $recording;
+
+    public function storeRecording(array $recording)
+    {
+        $this->recording = $recording;
+    }
+
+    public function next()
+    {
+        list ($this->position, $this->current) = each ($this->recording);
+
+        return $this->current;
+    }
+
+    public function valid()
+    {
+        return (boolean) $this->position;
+    }
+
+    public function rewind()
+    {
+        reset($this->recording);
+    }
+}


tests/VCR/Storage/AbstractStorageTest.php
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public function testFilePathCreated()
+    {
+        vfsStream::setup('test');
+        $this->storage = new TestStorage(vfsStream::url('test/foo'), 'file');
+    }
+
+}
+
+class TestStorage extends AbstractStorage
```
dbu: i hope its ok for you to have this class right in the test case file. imho it makes sense to keep them here, its "local" to just this test.

adri:
Yeah looks good. Are the tests working on your system? 
On travis they seem to fail https://travis-ci.org/php-vcr/php-vcr/jobs/51829624.

tests/VCR/Storage/AbstractStorageTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $this->storage = new TestStorage(vfsStream::url('test/'), 'folder/file');
+        $this->assertTrue($fs->hasChild('folder'));
+        $this->assertTrue($fs->getChild('folder')->hasChild('file'));
+    }
+
+    /**
+     * Expect error
+     */
+    public function testFilePathCreated()
```
dbu: todo: how can we test for this? does vfsStream provide anything?

src/VCR/Storage/AbstractStorage.php
```diff
@@ -50,12 +51,23 @@
...
+     *
      * @param string $cassettePath Path to the cassette directory.
-     * @param string $cassetteName Path to a file, will be created if not existing.
+     * @param string $cassetteName Path to the cassette file, relative to the path.
      */
     public function __construct($cassettePath, $cassetteName, $defaultContent = '[]')
     {
-        $file = $cassettePath . DIRECTORY_SEPARATOR . $cassetteName;
+        if (!is_dir($cassettePath)) {
+            throw new VCRException(sprintf('Cassette path "%s" is not existing or not a directory', $cassettePath), 0);
```
dbu: should i add a new code to VCRException? or use the global \InvalidArgumentException?

src/VCR/Storage/AbstractStorage.php
```diff
@@ -104,6 +116,8 @@ public function isNew() {
      */
     public function __destruct()
     {
-        fclose($this->handle);
+        if (is_resource($this->handle)) {
+            fclose($this->handle);
+        }
```
dbu: not directly related but i got very ugly console output when things go wrong in the Storage constructor.

tests/VCR/Storage/AbstractStorageTest.php
```diff
@@ -0,0 +1,64 @@
...
+    protected $storage;
+
+    public function testFilePathCreated()
+    {
+        $fs = vfsStream::setup('test');
+
+        $this->storage = new TestStorage(vfsStream::url('test/'), 'file');
+        $this->assertTrue($fs->hasChild('file'));
+
+        $this->storage = new TestStorage(vfsStream::url('test/'), 'folder/file');
```
dbu: this line is failing. do you see what i do wrong when creating the folder? it looks like i do the right thing and mkdir on vfsStream is not working properly... which seems odd.

dbu: ah. great. somebody claimed that i can just pass 0 to mkdir for permission mask, as that would trigger the default. but that seems to be wrong and the doc does not claim it would work. using 0777 things work now.

dbu:
okay, i am green locally.

src/VCR/Storage/AbstractStorage.php
adri: To me `Assertion::directory($cassettePath, "Cassette path "'{$cassettePath}" is not existing or not a directory");` would be most understandable, just like the other assertions a bit further down in the constructor. Also it throws the right exception.

The same check is btw done [here](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Configuration.php#L381) with a very descriptive error message intended for end users. But it certainly doesn't hurt to do it here as well.

src/VCR/Storage/AbstractStorage.php
```diff
@@ -50,12 +51,23 @@
...
+     * @param string $cassetteName Path to the cassette file, relative to the path.
      */
     public function __construct($cassettePath, $cassetteName, $defaultContent = '[]')
     {
-        $file = $cassettePath . DIRECTORY_SEPARATOR . $cassetteName;
+        if (!is_dir($cassettePath)) {
+            throw new VCRException(sprintf('Cassette path "%s" is not existing or not a directory', $cassettePath), 0);
+        }
+
+        $file = rtrim($cassettePath, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $cassetteName;
```
adri: It feels a bit like the constructor is doing too much. I would probably extend it to a method/helper. But its right "at the edge" so you might as well leave it like it is.

src/VCR/Storage/AbstractStorage.php
adri: Mh great find, thanks.

tests/VCR/Storage/AbstractStorageTest.php
adri: Mh I see, good to know :-)

tests/VCR/Storage/AbstractStorageTest.php
adri: Yeah its good, makes sense.

adri:
Its just the Assertion, otherwise I think its good. Thank you for this, cool feature.

src/VCR/Storage/AbstractStorage.php
dbu: oh, of course, sorry. i also suggest we keep it here - if people use a more IoC approach instead of going through the static factory, it will become useful. compared to the rest of the operations, a duplicate directory check should not be that expensive.

dbu:
okay, updated to use the assertion.

i had a look around and tried to see where i would document this but i can't find the right place to add this. it seems the whole cassette concept is never explained, only used in the examples. i suggest adding a section called "Usage" between "Installation" and "Configuration", and move the phpunit annotation that is in "Installation" to usage as well. then the usage can explain the main factory calls some more. shall i create an issue asking for that in the doc-repository? not sure if i find time to do that soon

src/VCR/Storage/AbstractStorage.php
adri: Yes definitely it should be here too. I agree that it is useful in any case.

I guess you're referring to the `VCR` static class. In general I'm not a fan of singletons for various reasons. 

Some background on the reasons for using a singleton in this case: With the current implementation of PHP-VCR I don't see how two Videorecorder instances can exist at the same time. The function calls (`curl`/`SoapClient::*`/...) are overwritten globally and VCR is either intercepting or not. Being able to have more instances would lead to strange bugs. 

Thinking about this... it could be actually possible to whitelist/blacklist some parts of code that does HTTP requests which should be intercepted and use it with one videorecorder instance and another part of code with another. Currently this is not implemented and I'm not sure if it is even worth it ;-)

adri:
Documentation is always nice :-) An issue would be good, but to be honest I don't know if I can get to it either. Thank you for this nice little feature!

adri:
Released https://github.com/php-vcr/php-vcr/releases/tag/1.2.3

src/VCR/Storage/AbstractStorage.php
```diff
@@ -50,12 +51,21 @@
...
+     * cassettePath are autocreated when not existing.
+     *
      * @param string $cassettePath Path to the cassette directory.
-     * @param string $cassetteName Path to a file, will be created if not existing.
+     * @param string $cassetteName Path to the cassette file, relative to the path.
      */
     public function __construct($cassettePath, $cassetteName, $defaultContent = '[]')
     {
-        $file = $cassettePath . DIRECTORY_SEPARATOR . $cassetteName;
+        Assertion::directory($cassettePath, 'Cassette path "{$cassettePath}" is not existing or not a directory');
```
adri: Ah wait I just saw the wrong usage of quotes here. I'll fix it on master

dbu:
created https://github.com/php-vcr/php-vcr.github.io/issues/6 about the doc

PR #110, opened 2015-02-26
Avoid writeable check when in mode 'none' to run on readonly filesystems

dbu:
On our CI, symfony sources are on a read-only filesystem and we certainly don't want php-vcr to update any fixtures. We switch to mode "none" based on an environment variable that is only present on CI. However, we would need this fix to avoid the writeable check (which makes no sense in "none" mode anyways).

/cc @thormeier

### DIFF
--- a/src/VCR/Configuration.php
+++ b/src/VCR/Configuration.php
@@ -121,16 +121,11 @@ class Configuration
     private $blackList = array('src/VCR/LibraryHooks/', 'src/VCR/Util/SoapClient', 'tests/VCR/Filter');
 
     /**
-     * The mode which determines how requests are handled
-     *
-     * Currently supported modes:
-     *      - new_episodes (Always allows new HTTP requests)
-     *      - once (Will allow new HTTP requests the first time the cassette is created then throw an exception after that)
-     *      - none (Will never allow new HTTP requests)
+     * The mode which determines how requests are handled. One of the MODE constants.
      *
      * @var string Current mode
      */
-    private $mode = 'new_episodes';
+    private $mode = VCR::MODE_NEW_EPISODES;
 
     /**
      * List of available modes.
@@ -143,9 +138,9 @@ class Configuration
      * @var array List of available modes.
      */
     private $availableModes = array(
-        'new_episodes',
-        'once',
-        'none'
+        VCR::MODE_NEW_EPISODES,
+        VCR::MODE_ONCE,
+        VCR::MODE_NONE,
     );
 
     /**
@@ -358,7 +353,7 @@ class Configuration
     /**
      * Sets the current mode.
      *
-     * @param string The mode to set VCR to
+     * @param string $mode The mode to set VCR to
      *
      * @return Configuration
      */
--- a/src/VCR/Storage/AbstractStorage.php
+++ b/src/VCR/Storage/AbstractStorage.php
@@ -54,30 +54,29 @@ abstract class AbstractStorage implements Storage
      * If the cassetteName contains PATH_SEPARATORs, subfolders of the
      * cassettePath are autocreated when not existing.
      *
-     * @param string $cassettePath Path to the cassette directory.
-     * @param string $cassetteName Path to the cassette file, relative to the path.
+     * @param string  $cassettePath   Path to the cassette directory.
+     * @param string  $cassetteName   Path to the cassette file, relative to the path.
+     * @param string  $defaultContent Default data for this cassette if its not existing
      */
     public function __construct($cassettePath, $cassetteName, $defaultContent = '[]')
     {
         Assertion::directory($cassettePath, "Cassette path '{$cassettePath}' is not existing or not a directory");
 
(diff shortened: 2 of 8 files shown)


src/VCR/Configuration.php
```diff
@@ -121,16 +121,26 @@ class Configuration
...
     /**
-     * The mode which determines how requests are handled
-     *
-     * Currently supported modes:
-     *      - new_episodes (Always allows new HTTP requests)
-     *      - once (Will allow new HTTP requests the first time the cassette is created then throw an exception after that)
-     *      - none (Will never allow new HTTP requests)
+     * Always allow to do HTTP requests and add to the cassette. Default mode.
+     */
+    const MODE_NEW_EPISODES = 'new_episodes';
```
dbu: constants is not strictly related to the fix, nor needed by it. but i think it would be better to use constants for this instead of handling around strings that can be mistyped.

dbu: uh. if you agree, i will replace all occurences of this with the constants in the php-vcr codebase e.g. in the Videorecorder.

src/VCR/Storage/AbstractStorage.php
```diff
@@ -54,10 +54,12 @@
...
      *
-     * @param string $cassettePath Path to the cassette directory.
-     * @param string $cassetteName Path to the cassette file, relative to the path.
+     * @param string  $cassettePath   Path to the cassette directory.
+     * @param string  $cassetteName   Path to the cassette file, relative to the path.
+     * @param string  $defaultContent Default data for this cassette if its not existing
+     * @param boolean $writeable      Whether the file must be writeable
      */
-    public function __construct($cassettePath, $cassetteName, $defaultContent = '[]')
+    public function __construct($cassettePath, $cassetteName, $defaultContent = '[]', $writeable = true)
```
dbu: i was pondering instead allowing to pass `false` for $defaultContent to indicate that this is readonly. as defaultContent makes no sense for readonly anyways. but that is a bit cryptic - on the other hand we now need to pass a default in the factory to get to the 4th parameter... what do you prefer? i tend to change it to say $defaultContent false means its readonly.

src/VCR/Storage/AbstractStorage.php
```diff
@@ -69,14 +71,15 @@ public function __construct($cassettePath, $cassetteName, $defaultContent = '[]'
 
         if (!file_exists($file) || 0 === filesize($file)) {
             file_put_contents($file, $defaultContent);
```
dbu: i don't check for $writeable in this case. this is to keep strict BC - if the file is missing and we provide a $defaultContent or there are no requests, there is no error.

on a readonly filesystem, the file_put_contents will fail however. but i guess we don't care about that, its not different than before.

src/VCR/Storage/AbstractStorage.php
```diff
@@ -69,14 +71,15 @@ public function __construct($cassettePath, $cassetteName, $defaultContent = '[]'
...
-
             $this->isNew = true;
+        } else {
+            Assertion::file($file, "Specified path '{$file}' is not a file.");
+            Assertion::readable($file, "Specified file '{$file}' must be readable.");
+            if ($writeable) {
+                Assertion::writeable($file, "Specified path '{$file}' must be writeable.");
+            }
```
dbu: no need to do these checks if we just created the file - obviously it was writeable and readable or file_put_contents would have failed.

src/VCR/Configuration.php
adri: Good idea! 

I'm a bit torn where to put the constants. Either on the configuration or directly on the `VCR` class. 

I wanted to provide a nicer interface for adding [events](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/VCREvents.php) as well and was thinking about putting it all on the `VCR` object. That would look like this:

```
VCR::configuration()
    ->setMode(VCR::MODE_NEW_EPISODES)
    ->on(VCR::BEFORE_RECORD, function() {...})
```

What would you like more as a user? `Configuration` or `VCR`. I'm more in favour of `VCR` right now because using only a single interface seems kind of neat.

src/VCR/Storage/AbstractStorage.php
adri: Is there really a need to introduce another variable? I would have moved the writeable check down to the `storeRecording` where stuff gets written. The downside is that the check would have to be added in the `Json` and `Yaml` storage. Am I missing something?

dbu: maybe the better idea, especially as the yaml constructor follows a different pattern (causing the failures in travis).

this delays reporting the error to when writing is actually attempted but whatever, its just testing so you are supposed to look at the output. i will change to that

dbu:
thanks for the feedback. i updated the code, constants on VCR make more sense indeed.

once this is released, we should also update the documentation to use the constants.

dbu:
thanks @adri - can you tag another release with this?

adri:
Done already, forgot to post it here https://github.com/php-vcr/php-vcr/releases/tag/1.2.4

dbu:
great, thanks!

Issue #94, opened 2015-01-19
Feature request: indicate why cassette does not have necessary response


tylercollier:
I got this message:

```
LogicException: Invalid http request. The cassette inserted did not have the necessary response. If you want to send a request anyway, make sure your mode is set to new_episodes.
```

Another developer had modified our code, such that the body that was used for an http request was different. He did not modify the cassette, which he should have.

My issue is with the message. How about something like "The request body does not match a previous request"? Or, as I was digging through the code, the failure was not actually on the body, but the `Content-Length`. Same difference... I think a message saying "The request header ("Content-Length: 1567") did not match an existing request" would be more helpful.

This took me an hour to figure out myself, because I had to dig through the PHP-VCR source code. A more helpful message would've helped me understand the issue immediately.

tylercollier:
As follow-up motivation, I have another developer running tests, and they are failing with that same message. Helping him debug this remotely is going to be a pain. The fact that he's a rookie is not the fault of php-vcr, but why doesn't it just say

    Invalid http request because [REASON]

?

adri:
That you very much Tyler. I think so to there should be a way to easier find out what is happening (#36 is related to this as well). 

"The request header ("Content-Length: 1567") did not match an existing request": How can we find out which parts request did not match? Not sure if it is possible. What might work is a list of all requests in a cassette with information why it didn't match.

I'm thinking about using #86 to provide a `debug` mode which then logs information somewhere. I wonder where that ideally should be. It could be in the exception as part of the message:  

```
Invalid HTTP request. [... more information about the request ...] Recorded requests in the cassette:
 * GET /example, didn't match the header
 * GET /example2, didn't match the body
```

What do you think?

tylercollier:
It's easy to find out which parts didn't match a given request, right? It has to be; that's how VCR works. I assume what you mean is, how to convey that information to the user without overwhelming them, if there are multiple requests in a cassette? One idea is to only show what part didn't match if there's a single request in the given cassette. That works well for me, because I only ever have a single request per cassette. Even then, if it didn't match the header, I'd like to see why. Show me the current request and the one from the cassette. Any example where this works in (non-PHP) VCR is [here](http://myronmars.to/n/dev-blog/2012/02/vcr-2-0-0-rc2-released): you can set the debug logger during bootstrapping.

[...]

PR #111, opened 2015-03-02
Mismatch explanations

tylercollier:
This is for #94.

I said I'd have this done by tonight. It's not fully finished, but close. I'm submitting for your comments.

I'm still working on the last piece which is ["detect the recorded request which looks like the most to the request"](https://github.com/php-vcr/php-vcr/issues/94#issuecomment-71310952). Please see my proposed algorithm in `MatchObserver.php`, in the comments.

The 'base'/default set of request matchers are now implemented as classes, but one can still use simple callables.

I wrote tests, and tested them each, but I didn't run the rest of the test suite so I'm sure I broke some. I will fix them.


src/VCR/RequestMatcher.php
```diff
@@ -1,102 +0,0 @@
-<?php
-
-namespace VCR;
-
-/**
- * Collection of matcher methods to match two requests.
- */
-class RequestMatcher
```
dbu: is there a risk of BC breaks when removing this class? or is it only ever used internally?

tylercollier: I had to look up ["BC breaks"](http://stackoverflow.com/q/8891005/135101) :-)

You're right. Those methods are public so of course they could be used externally by anyone. We could leave them in indefinitely, and optionally put deprecation warnings.

src/VCR/Configuration.php
```diff
@@ -14,6 +14,11 @@
  */
 class Configuration
 {
+    public function __construct() {
+        $this->matchObserver = new MatchObserver();
```
adri: The configuration is currently a simple container for configuration data. All instance creation should happen in the factory and then handed in where it is needed, see "inversion of control".

When instance creation logic moves to the `VCRFactory` the `matcherObserver` should not be needed in the configuration.

src/VCR/MatchObserver.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace VCR;
+
+class MatchObserver
+{
+    // This is an array keyed by a hash of the (first aka stored) request.
+    private $requestToMatchersMap;
+    private $requestCount0;
```
adri: I'm a bit confused with all the variables, `requestToRequestsMap`, `requestCountMap`, `requestToMatcherMap`.

`shouldObserver`: I wonder if we should provide a way to disable it. After all we are normally in a 'testing' mode and it would be very useful to always have that information. Performance I guess shouldn't be a big problem since it is not meant for production and only happens when a request does not match.

src/VCR/MatchObserver.php
```diff
@@ -0,0 +1,83 @@
...
+
+    public function clear() {
+        $this->requestToMatchersMap = array();
+        $this->requestCount = 0;
+        $this->requestToCountMap = array();
+        $this->requestToRequestsMap = array();
+    }
+
+    public function markMismatch(Request $first, Request $second, $matcher) {
+        $hash = $first->getIdentityHash();
```
adri: Whats the idea with the identity hash?

src/VCR/MatchObserver.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace VCR;
+
+class MatchObserver
```
adri: At the moment the MatchObserver is a singleton. It could also be a value object which gets filled with data and then thrown away.

src/VCR/Videorecorder.php
```diff
@@ -216,7 +216,10 @@ public function handleRequest(Request $request)
             );
         }
 
+        $this->getMatchObserver()->startObserving();
+
         if ($this->cassette->hasResponse($request)) {
+            $this->getMatchObserver()->stopObserving();
```
adri: I have the feeling that `hasResponse` and `playback` are conceptually something else than finding out why a request doesn't match. Now it's doing more than one thing (see 'single responsibility principle'). 

How about introducing another method on the `Cassette` which does what you want and is named like that. Like `getMismatchReport()` or `getMismatchInfo()`. I imagine it would return an object like the `MatchObserver` which can be asked for a `getMismatchMessage()`.

src/VCR/Cassette.php
```diff
@@ -68,6 +68,10 @@ public function hasResponse(Request $request)
      */
     public function playback(Request $request)
     {
+        return $this->getResponse($request);
+    }
+
+    public function getResponse(Request $request) {
```
adri: Should be at least `protected`.

src/VCR/RequestMatcher.php
adri: I think the BC break is that the request matchers changed. They used to be `callable`s and now are objects. Doing a `addRequestMatcher` and using that with `enableRequestMatchers` in the current way is not possible http://php-vcr.github.io/documentation/configuration/.

src/VCR/MatchObserver.php
adri: Not sure about the `MatchObserver` name. When I read observer I think of the [observer pattern](http://en.wikipedia.org/wiki/Observer_pattern) which isn't really the case here. 

To me it seems like the purpose of that class is to explain mismatches. Maybe `MismatchExplainer`? ;-)

adri:
I made some edits to my comments, not sure if you see that with email. Hope it makes sense.

src/VCR/Cassette.php
tylercollier: Agreed

src/VCR/Configuration.php
tylercollier: I did not pay careful enough attention to the factory. Cool, I will move logic to there.

src/VCR/MatchObserver.php
tylercollier: I'm embarassed on this code. I'm sure there are better ways to organize it that are simpler and more straightforward conceptually, and also much easier to read! I will clean it up.

I agree on the should observe. The problem I had was that multiple calls to getRequest (first hasResponse uses it, then playback uses it) would record multiple mismatches. By changing the mismatch explainer to be a value / not a service, this problem will be removed: we won't need a should observe / start observing / stop observing.

src/VCR/MatchObserver.php
tylercollier: I didn't want the order of matching to matter, so an identity hash was a way to keep matches with their request.

src/VCR/MatchObserver.php
tylercollier: You're correct that it's a service. I don't like that. I will look into making it a value object.

I agree with you on the name. I originally was thinking that the observer pattern would apply here, and kept that name. I like your name.

src/VCR/RequestMatcher.php
tylercollier: I added classes for the matchers, but what I put in the availableRequestMatchers member (in the createAvailableMatchers method) are still callables. You can still use addRequestMatcher and enableRequestMatchers as before. I just ran `phpunit ./tests/VCR/ConfigurationTest.php`:

```
There was 1 failure:
1) VCR\ConfigurationTest::testEnableRequestMatchers
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
-        0 => 'VCR\RequestMatcher'
-        1 => 'matchHeaders'
+        0 => VCR\RequestMatchers\HeadersMatcher Object (...)
+        1 => 'match'
     )
     1 => Array (
-        0 => 'VCR\RequestMatcher'
-        1 => 'matchBody'
+        0 => VCR\RequestMatchers\BodyMatcher Object (...)
+        1 => 'match'
     )
 )
/home/tylercollier/repos/php-vcr-tylercollier/tests/VCR/ConfigurationTest.php:80
```

Simple to fix up that test :-)

Just for own my own double-check, I put this in `ConfigurationTest.php`, and it passes:

```
public function returnTrue() {
    return true;
}

public function testAddRequestMatchersWithObjectAndMethodNameCallable()
{
    $expected = array(&$this, 'returnTrue');
    $this->config->addRequestMatcher('new_matcher', $expected);
    $this->assertContains($expected, $this->config->getRequestMatchers());
}
```

@dbu is correct though that if people had referenced one of the static functions (e.g. matchMethod), then maybe I should put that class back. I would recommend deprecating them, since we have improved replacements.

src/VCR/Videorecorder.php
tylercollier: I agree. I didn't want to "mess too much" with the existing code/order, but I think that was my mistake as a novice open-source contributor. What I ended up doing made less sense. I will use your idea.

tylercollier:
I'm not sure how to know if I saw your original comments or the edited ones. Let me know if I answered what you wanted.

I will have my next iteration done by the end of March 8 GMT-0700.

src/VCR/MatchObserver.php
adri: You could also use the [SplObjectStorage](http://php.net/manual/en/class.splobjectstorage.php) for that, which doesn't require adding the `getIdentityHash()` function on the object.

adri:
Some feedback on the general architecture:

Right now the output to the user is spread across multiple classes. I think it would be better if everything related to building the message for the user is put in a `MismatchExplainer`/`MatchObserver` (which `MismatchExplainer` to use could be configured if there should be something more advanced). Since the mismatches are explained per request it can take the $request in the constructor `MismatchExplainer($request)`.

I think the properties on the `MismatchExplainer ` could be made clearer if you would exact this information in separate objects. There might be for example a `RequestMismatch(Request $request, FieldMismatch[] $fieldMismatches)` object which keeps track of mismatches for each request in the cassette. Not sure if `FieldMismatch` is the best name though ;-)

A strategy to find the best match can then be to loop over the list of `RequestMismatch` objects and return the one with the minimum `FieldMismatch`es, but more than 0. The index of the best match is then the index of the `RequestMismatch` object in the list. 

Those `FieldMismatch($fieldName, $firstValue, $secondValue)`es could be returned from the `RequestMatcher`s because they have the necessary information to create those objects. Right now their `match` method returns a `boolean`. So this would need some rework...

I quite like that you extracted the `RequestMatcher`s  to their own classes, they could get another method which returns a list of `FieldMismatch`es (like `getFieldMismatches`). To make this work properly I think the current callable `RequestMatcher`s should be reworked into using proper classes with an intercase like you already wrote. 

To keep compatibility with existing `RequestMatchers` we could convert a `callable` in `Configuration->addRequestMatcher()` to a `CustomRequestMatcher($name, $callable)`. 

Please let me know your thoughts on that or if you have any questions.

src/VCR/MatchObserver.php
tylercollier: Good find!

tylercollier:
I think I understand your ideas. Putting the output in a single place is a good idea. I did it how I did before because I didn't want to "rock the boat" too much with how things are done. I realize how that turned out to be a mistake.

The code would be quite readable and extendable with `RequestMismatch` objects. `FieldMismatch` is not a bad name, but Field is used a lot with the term "header field" in [the spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html). "Part" makes me think of "multipart". There is no "rollup" term [used here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4) that covers what we need. So I think Field is fine for now.

tylercollier:
I did not finish this tonight like I said I would. I've pushed a commit solely for the sake of review. Don't bother running tests.

I want to think more about RequestMismatch and FieldMismatch. I admit I don't understand what they'd contain, especially how FieldMismatch is different than RequestMatcher (it seems like FieldMismatch's job would just be to have a RequestMatcher). Is RequestMismatch really meant to hold mismatches for all requests of a cassette, or just one, and then MismatchExplainer would have one RequestMismatch for each request?

Changes:
- Rename MatchObserver to MismatchExplainer
- Remove constructor on Configuration.php, add createVCRConfiguration in VCRFactory

src/VCR/Request.php
```diff
@@ -75,6 +75,13 @@ public function matches(Request $request, array $requestMatchers)
         return true;
     }
 
+    public function collectMismatches(Request $request, array $requestMatchers, MismatchExplainer $mismatchExplainer) {
+        // convert simple callables here to CallableRequestMatcher
```
adri: I was thinking of doing that way in the beginning when calling `Configuration::addRequestMatcher`. So here `collectMismatches` would get a list of RequestMatcher objects which have a `match` and a `getMismatches` method.

adri:
Don't worry, thank you very much for your time and work!

The way I imagined it, `RequestMismatch` would indeed have all mismatches for **one request**.  
The `MismatchExplainer` operates on a list of `RequestMismatch`es, one for each request in the cassette. Each `RequestMismatch` has a list of (field) mismatches. That way I imagine the code could be pretty readable and extendable.

The difference between a `RequestMatcher` and "field" mismatch is that a FieldMismatch object is just a value object containing information what doesn't match. It would be returned from a `RequestMatcher`.

```php
// HostMatcher implements RequestMatcher
public function getMismatches(Request $first, Request $second) {
    if ($this->matches($first, $second)) {
        return array();
    }

    return array(
        new FieldMismatch('host', $first->getHost(), $second->getHost())
    );
}
```

A `HeaderMatcher` for example might return a list of many mismatches if there are multiple header fields not matching. All mismatches of all used request matchers would be merged and stored on a RequestMatch.

I see the following concepts:

 * Current request "$first"
 * Stored request  "$second"
 * `RequestMatcher` matches two requests for one or more specific fields and returns `FieldMismatch`es.
 * `RequestMatcher\FieldMismatch($field, $firstValue, $secondValue)` stores information about a single mismatch.
 * `RequestMatcher\RequestMismatch($request, array $fieldMismatches)` stores all mismatches for one request.
 * `RequestMatcher\MismatchExplainer($request, $requestMismatchList)` explain mismatches for requests in a readable form.
 * `Cassette::collectMismatches($request, $requestMatchers)` loops requests in a cassette and merges all mismatches returned by the request matchers. 

You could do this also without intermediary objects like `FieldMismatch` and `RequestMismatch` by using simple arrays and maps. The downside to that is, that it could end up being less explicit and less extendable. We won't need the

Issue #101, opened 2015-02-05
BeforePlaybackEvent does not happen before playback


tylercollier:
Kudos to @aaa2000 who added hooks on #86.

The BeforePlaybackEvent does not actually happen before playback. Look at `Videorecorder.php`:

```
public function handleRequest(Request $request)
{
    ...
    if ($this->cassette->hasResponse($request)) {
        $this->dispatch(VCREvents::VCR_BEFORE_PLAYBACK, new BeforePlaybackEvent($request, $this->cassette));
        $response = $this->cassette->playback($request);
        $this->dispatch(VCREvents::VCR_AFTER_PLAYBACK, new AfterPlaybackEvent($request, $response, $this->cassette));
```

`hasResponse` just calls `playback`. Thus, the code should look more like this:

```
$this->dispatch(VCREvents::VCR_BEFORE_PLAYBACK, new BeforePlaybackEvent($request, $this->cassette));
if ($this->cassette->hasResponse($request)) {
    $response = $this->cassette->playback($request);
    $this->dispatch(VCREvents::VCR_AFTER_PLAYBACK, new AfterPlaybackEvent($request, $response, $this->cassette));
    ...
```

This feels gross, but is more correct.

So, how to fix the grossness of dispatching the event before the `if`? Some ideas:
* Dispatch the VCR_BEFORE_PLAYBACK event in the actual `playback` method. This means the event would be dispatched each time playback happens. That might be what's wanted. If not, we could keep track if the event has already been dispatched.
* Why is `playback` responsible for finding a response? Doesn't that make more sense for `hasResponse` to do? Or really, I think the code that does the work of finding find the response should be moved from `playback` into a `getResponse` method, which is called both from `hasResponse` and `playback`.

I'm happy to do any of this work. Just asking for input first.

tylercollier:
Here's my motivation if it matters.

I am using Guzzle, which sets the User-Agent header to be something like

    User-Agent: 'Guzzle/3.9.2 curl/7.22.0 PHP/5.4.23'

The problem with this is that that's MY version of curl and PHP. If another developer has a different version, the request won't match. This scenario is happening on my team.

So my VCR_BEFORE_PLAYBACK hook strips the User-Agent header if it finds version numbers in it.

But this is why the BEFORE_PLAYBACK hook must actually run earlier than it currently does.

adri:
Yes you are right, good point. We could move the content of the if above and then check the response. 
[...]

PR #118, opened 2015-06-04
Dispatch before playback event for each http interaction in cassette

tylercollier:
Fix for #101.

As I mention [in this comment](https://github.com/php-vcr/php-vcr/issues/101#issuecomment-108513146), I followed Ruby VCR's lead (and aaa2000's suggestion) and dispatch the "before playback" event before each HTTP interaction in the cassette.

I did not do change the 'after playback' event, since there should be one anyway.

I did end up moving the dispatch into the configuration, a la Ruby's VCR. This means the `dispatch` method needs to be public.

### DIFF
--- a/src/VCR/Cassette.php
+++ b/src/VCR/Cassette.php
@@ -4,6 +4,7 @@ namespace VCR;
 
 use VCR\Storage\Storage;
 use VCR\Util\Assertion;
+use VCR\Event\BeforePlaybackEvent;
 
 /**
  * A Cassette records and plays back pairs of Requests and Responses in a Storage.
@@ -56,7 +57,7 @@ class Cassette
      */
     public function hasResponse(Request $request)
     {
-        return $this->playback($request) !== null;
+        return $this->getResponse($request, true) !== null;
     }
 
     /**
@@ -68,8 +69,24 @@ class Cassette
      */
     public function playback(Request $request)
     {
+        return $this->getResponse($request);
+    }
+
+    /**
+     * Returns a response for given request or null if not found.
+     * Optionally dispatches VCR_BEFORE_PLAYBACK event.
+     *
+     * @param Request $request Request to check if it was recorded.
+     * @param boolean $dispatchEvents Whether to dispatch events or not. Default is false.
+     *
+     * @return boolean True if a response was recorded for specified request.
+     */
+    protected function getResponse(Request $request, $dispatchEvents = false) {
         foreach ($this->storage as $recording) {
             $storedRequest = Request::fromArray($recording['request']);
+            if ($dispatchEvents) {
+                $this->config->dispatch(VCREvents::VCR_BEFORE_PLAYBACK, new BeforePlaybackEvent($request, $this));
+            }
             if ($storedRequest->matches($request, $this->getRequestMatchers())) {
                 return Response::fromArray($recording['response']);
             }
--- a/src/VCR/Cassette.php
+++ b/src/VCR/Cassette.php
@@ -79,13 +79,13 @@ class Cassette
      * @param Request $request Request to check if it was recorded.
      * @param boolean $dispatchEvents Whether to dispatch events or not. Default is false.
      *
-     * @return boolean True if a response was recorded for specified request.
+     * @return Response|null Response for specified request.
      */
     protected function getResponse(Request $request, $dispatchEvents = false) {
         foreach ($this->storage as $recording) {
             $storedRequest = Request::fromArray($recording['request']);
             if ($dispatchEvents) {
-                $this->config->dispatch(VCREvents::VCR_BEFORE_PLAYBACK, new BeforePlaybackEvent($request, $this));
+                $this->config->dispatch(VCREvents::VCR_BEFORE_PLAYBACK, new BeforePlaybackEvent($storedRequest, $this));
             }
             if ($storedRequest->matches($request, $this->getRequestMatchers())) {
                 return Response::fromArray($recording['response']);
--- a/src/VCR/Configuration.php
+++ b/src/VCR/Configuration.php
@@ -3,6 +3,9 @@
 namespace VCR;
 
 use VCR\Util\Assertion;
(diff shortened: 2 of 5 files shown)


src/VCR/Cassette.php
```diff
@@ -68,8 +69,24 @@ public function hasResponse(Request $request)
...
+     * @param Request $request Request to check if it was recorded.
+     * @param boolean $dispatchEvents Whether to dispatch events or not. Default is false.
+     *
+     * @return boolean True if a response was recorded for specified request.
+     */
+    protected function getResponse(Request $request, $dispatchEvents = false) {
         foreach ($this->storage as $recording) {
             $storedRequest = Request::fromArray($recording['request']);
+            if ($dispatchEvents) {
+                $this->config->dispatch(VCREvents::VCR_BEFORE_PLAYBACK, new BeforePlaybackEvent($request, $this));
```
aaa2000: I think that the request of the BeforePlaybackEvent should be the storedRequest.

src/VCR/Cassette.php
```diff
@@ -68,8 +69,24 @@ public function hasResponse(Request $request)
...
+    }
+
+    /**
+     * Returns a response for given request or null if not found.
+     * Optionally dispatches VCR_BEFORE_PLAYBACK event.
+     *
+     * @param Request $request Request to check if it was recorded.
+     * @param boolean $dispatchEvents Whether to dispatch events or not. Default is false.
+     *
+     * @return boolean True if a response was recorded for specified request.
```
aaa2000: @return Response|null

src/VCR/Cassette.php
tylercollier: Yes, thanks!

src/VCR/Cassette.php
tylercollier: I'm sorry, I'm not clear on what you're telling me here.

aaa2000: The phpdoc is incorrect.  
```
* @return boolean True if a response was recorded for specified request
```
instead of
```
* @return  Response|null Returns a response if a response was recorded for specified request
```

tylercollier: Ah ha. Thanks!

adri:
Thanks. I think its all good except that dispatching events happens in the configuration. Is this only for convenience? Why not create a event dispatcher class which is injected into the classes which need it?
I don't think that dispatching events and configuration belong together.

adri:
Well it's ok for now. Can you please rebase?

tylercollier:
@adri, I dispatched from configuration for convenience, like you said. I believe this is how I saw it in Ruby.

When you ask me to rebase, do you mean squash my two commits into one? Or do you mean rebase the branch onto master?

adri:
Hey Tyler,

Ah yeah I meant rebasing the branch on master. At the moment I can't merge automatically :-)
I meanwhile fixed the test setup, the PHP 7 tests failures don't let the build fail now. So we can see after the rebase if everything builds nicely.

tylercollier:
You already merged aaa2000's pull request, right? My pull request is incompatible. We wouldn't want both implementations. I gave you push access to my fork in case I'm misunderstanding and you want to take care of it quickly yourself.

adri:
Ok I understand. Should the `VCR_BEFORE_PLAYBACK` event really be issued for each stored request in the cassette? I would expect it is called only once with just a request like in @aaa2000 implementation.

adri:
In Ruby it is https://www.relishapp.com/vcr/vcr/v/2-9-1/docs/hooks/before-playback-hook

> Your block should accept up to 2 arguments. The first argument will be
the HTTP interaction that is about to be used for play back. The second
argument will be the current cassette.

I read it as it is issuing only one event, am I wrong?

tylercollier:
It's up to us I suppose, but that's how it's done in Ruby's VCR. I was just trusting that they did that for a good reason, and not an arbitrary one. But I admit I didn't do research to verify, and nothing jumps out at me as to do it one way or the other.

> I read it as it is issuing only one event, am I wrong?

I think this sentence: `The first argument will be the HTTP interaction that is about to be used for play back.` means it will be for each request in the cassette. See [the code](https://github.com/vcr/vcr/blob/master/lib/vcr/cassette.rb#L176).

aaa2000:
I prefer @tylercollier implementation because my PR dispatches the VCR_BEFORE_PLAYBACK and VCR_AFTER_PLAYBACK event when the cassette has no response.

tylercollier:
@adri Did you have a chance to look at this?

We just implemented a deployment strategy which requires all tests to pass before it will deploy, so my co-workers and I are depending on this change. We will use it to ignore HTTP headers with PHP/Curl version differences.

adri:
Not yet unfortunately.

>  We will use it to ignore HTTP headers with PHP/Curl version differences.
This should be possible with using custom request matchers. You can disable the `headers` request matcher and add your own if you would like to test for some specific headers. See http://php-vcr.github.io/documentation/configuration/.

tylercollier:
Here's what I did in the meantime:

```
// Each developer will have different versions of PHP, Curl, and
// perhaps other libraries. The library versions are often included in API
// requests, meaning that VCR would create a cassette for one developer
// differently than another, at least for the User-Agent header. To resolve
// this, we'll remove the User-Agent header.
\VCR\VCR::configure()->addRequestMatcher(
	'modified_headers_matcher',
	function (\VCR\Request $first, \VCR\Request $second) {
		// In the following, we are blacklisting the 'User-Agent' key.
		$firstHeaders = array_diff_key($first->getHeaders(), array_flip(['User-Agent']));
		$secondHeaders = array_diff_key($second->getHeaders(), array_flip(['User-Agent']));
		return array_filter($firstHeaders) === array_filter($secondHeaders);
	}
);
\VCR\VCR::configure()->enableRequestMatchers(['method', 'url', 'host', 'body', 'post_fields', 'query_string', 'modified_headers_matcher']);
```

adri:
Closing in favour of #127.

Issue #120, opened 2015-06-13
PHP VCR not sending request body


justincy:
When PHP VCR is enabled the request bodies don't get sent on a POST even though the `Content-Length` header is set and it correctly records the body.

justincy:
I can't figure out what's wrong from debugging and examining the source code. However I do see that POST is not actually tested in integration tests.

adri:
Some questions: 
What is the content of the POST body when you send it? 
What HTTP library are you using?

adri:
Thank you Justin for fixing this in #122. Is your original issue solved?

justincy:
Yes. Thanks.

Issue #103, opened 2015-02-19
Error when checking for casette file is_writable in vagrant


dbu:
we get the following exception:

```
VCR\VCRException: Specified path 'tests/fixtures/default' must be writeable.

/var/www/vendor/beberlei/assert/lib/Assert/Assertion.php:205
/var/www/vendor/beberlei/assert/lib/Assert/Assertion.php:1177
/var/www/vendor/php-vcr/php-vcr/src/VCR/Storage/AbstractStorage.php:68
/var/www/vendor/php-vcr/php-vcr/src/VCR/Storage/Yaml.php:36
/var/www/vendor/php-vcr/php-vcr/src/VCR/VCRFactory.php:53
/var/www/vendor/php-vcr/php-vcr/src/VCR/VCRFactory.php:119
/var/www/vendor/php-vcr/php-vcr/src/VCR/VCRFactory.php:98
/var/www/vendor/php-vcr/php-vcr/src/VCR/Videorecorder.php:177
/var/www/vendor/php-vcr/php-vcr/src/VCR/VCR.php:20
/var/www/src/Acme/Tests/MyTestCase.php:42
```

the file is created with 664, and when typing `ls -l` the system reports the file as belonging to user 1000 (which is my user id on the host system). inside vagrant, the vagrant user has id 501. but somehow the linux filesystem can still write the file. and the is_writable function is true when the stream wrapper is not registered. as soon as its registered, files are considered not writable. 

i managed to confirm the problem by changing the vagrant user id inside the box to user 1000. `ls -l` reports the file as owned by the "vagrant" user instead of numeric now. with this, the file is considered writable even with the stream wrapper.

i could live with the workaround except for our continuous integration that is seeing the same problem.

Danno040:
I believe I encountered the same issue as @dbu. When I called `VCR::turnOn`, files that were writable were starting to register as not writable.

```php
echo var_export(is_writeable('/dev-env/website/storage/cms/twig/e/4')); // Returns true
require './bootstrap/autoload.php';
echo var_export(is_writeable('/dev-env/website/storage/cms/twig/e/4')); // Returns true
\VCR\VCR::turnOn();
echo var_export(is_writeable('/dev-env/website/storage/cms/twig/e/4')); // Returns false
```

I believe this has something to do with the weird permissions that are set when using NFS with vagrant. Using the [vagrant-bindfs](https://github.com/gael-ian/vagrant-bindfs) plugin fixed the issue for me.

[...]

PR #122, opened 2015-06-13
Fix POST requests

justincy:
Here are failing test cases for #120. The tests seem to be exhibiting a different, though perhaps related, problem than what I was seeing. I'll have to debug further to figure out what's going wrong. Any ideas?

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -175,6 +175,7 @@ class CurlHelper
             case CURLOPT_HEADER:
             case CURLOPT_WRITEFUNCTION:
             case CURLOPT_HEADERFUNCTION:
+            case CURLOPT_UPLOAD:
                 // Ignore header, file and writer functions.
                 // These options are stored and will be handled later in handleOutput().
                 break;
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -179,21 +179,34 @@ class CurlHelper
                 // Ignore header, file and writer functions.
                 // These options are stored and will be handled later in handleOutput().
                 break;
-            case CURLOPT_READFUNCTION:
-                // Guzzle provides a callback to let curl read the body string.
-                // To get the body, this callback is called manually.
-                if (is_null($value)) {
-                    $request->setCurlOption($option, $value);
-                    break;
-                }
-                $bodySize = $request->getCurlOption(CURLOPT_INFILESIZE);
-                Assertion::notEmpty($bodySize, "To set a CURLOPT_READFUNCTION, CURLOPT_INFILESIZE must be set.");
-                $body = call_user_func_array($value, array($curlHandle, fopen('php://memory', 'r'), $bodySize));
-                $request->setBody($body);
-                break;
             default:
                 $request->setCurlOption($option, $value);
                 break;
         }
+    }
+    
+    /**
+     * Makes sure we've properly handled the POST body, such as ensuring that
+     * CURLOPT_INFILESIZE is set if CURLOPT_READFUNCTION is set.
+     *
+     * @param Request  $request Request to set cURL option to.
+     */
+    public function validateCurlPOSTBody(Request $request)
+    {
+        $readFunction = $request->getCurlOption(CURLOPT_READFUNCTION);
+        if (is_null($readFunction)) {
+            return;
+        }
+        
+        // Guzzle 4 sometimes sets the post body in CURLOPT_POSTFIELDS even if
+        // they have already set CURLOPT_READFUNCTION.
+        if ($request->getBody()){
+            return;
+        }
+        
+        $bodySize = $request->getCurlOption(CURLOPT_INFILESIZE);
+        Assertion::notEmpty($bodySize, "To set a CURLOPT_READFUNCTION, CURLOPT_INFILESIZE must be set.");
+        $body = call_user_func_array($readFunction, array($curlHandle, fopen('php://memory', 'r'), $bodySize));
+        $request->setBody($body);
     }
 }
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -191,7 +191,7 @@ class CurlHelper
      *
      * @param Request  $request Request to set cURL option to.
      */
-    public function validateCurlPOSTBody(Request $request)
+    public static function validateCurlPOSTBody(Request $request)
     {
         $readFunction = $request->getCurlOption(CURLOPT_READFUNCTION);
         if (is_null($readFunction)) {
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -190,8 +190,9 @@ class CurlHelper
      * CURLOPT_INFILESIZE is set if CURLOPT_READFUNCTION is set.
      *
      * @param Request  $request Request to set cURL option to.
+     * @param resource $curlHandle cURL handle associated with the request.
      */
-    public static function validateCurlPOSTBody(Request $request)
+    public static function validateCurlPOSTBody(Request $request, $curlHandle = null)
     {
         $readFunction = $request->getCurlOption(CURLOPT_READFUNCTION);
         if (is_null($readFunction)) {
--- a/tests/integration/guzzle/3/ExampleHttpClient.php
+++ b/tests/integration/guzzle/3/ExampleHttpClient.php
@@ -24,6 +24,25 @@ class ExampleHttpClient
 
         return null;
     }
+    
+    public function post($url, $body)
+    {
+        $client = new Client();
+        
+        try {
+            $request = $client->post($url);
+            $request->setBody($body);
+            $response = $request->send();
+
+            return json_decode($response->getBody(), true);
+        } catch (ClientErrorResponseException $e) {
+            if ($e->getResponse()->getStatusCode() !== 404) {
+                throw $e;
+            }
+        }
+
+        return null;
+    }
 }
 
 
--- a/tests/integration/guzzle/3/ExampleHttpClient.php
+++ b/tests/integration/guzzle/3/ExampleHttpClient.php
@@ -31,7 +31,7 @@ class ExampleHttpClient
         
         try {
             $request = $client->post($url);
-            $request->setBody($body);
+            $request->setBody($body, 'application/json');
             $response = $request->send();
 
             return json_decode($response->getBody(), true);
--- a/tests/integration/guzzle/4/ExampleHttpClient.php
+++ b/tests/integration/guzzle/4/ExampleHttpClient.php
@@ -28,7 +28,7 @@ class ExampleHttpClient
         $client = new Client();
         
         try {
-            $response = $client->post($url, ['body' => $body]);
+            $response = $client->post($url, array('body' => $body));
             return $response->json();
         } catch (ClientErrorResponseException $e) {
             if ($e->getCode() !== 404) {
--- a/tests/integration/guzzle/4/ExampleHttpClient.php
+++ b/tests/integration/guzzle/4/ExampleHttpClient.php
@@ -22,4 +22,20 @@ class ExampleHttpClient
 
         return null;
     }
+
+    public function post($url, $body)
+    {
+        $client = new Client();
+        
+        try {
+            $response = $client->post($url, ['body' => $body]);
+            return $response->json();
+        } catch (ClientErrorResponseException $e) {
+            if ($e->getCode() !== 404) {
+                throw $e;
+            }
+        }
+
+        return null;
+    }
 }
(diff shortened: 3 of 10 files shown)


adri:
Thank you very much for you test cases!

adri:
I looked into it and it seems to be an issue with the virtual file system. When I change the test `setUp` method to use `\VCR\VCR::configure()->setCassettePath('/tmp');` the tests run fine.

adri:
From the recorded fixture file:

```
    request:
        method: POST
        url: 'http://httpbin.org/post'
        headers:
            Host: httpbin.org
            Expect: null
            Content-Type: null
            Accept-Encoding: null
            User-Agent: 'Guzzle/5.1.0 curl/7.38.0 PHP/5.5.20'
            Accept: null
        body: foo=bar
```

justincy:
@adri That change doesn't fix the tests for me. Maybe it's a problem with a specific PHP + curl version. I'm using PHP 5.5.9 with curl version 7.35.0.

justincy:
@adri Removing vfs fixes Guzzle 5 but not 4 or 3.

adri:
Thanks for the info. I noticed that the error message is different for [Guzzle 3](https://travis-ci.org/php-vcr/php-vcr/jobs/66870715).

adri:
Seems like in Guzzle version 3 there is some CURL option set which leads to the the following error [HttpClient](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Util/HttpClient.php#L34
), `curl_error($ch)` returns:
`Recv failure: Connection reset by peer`

adri:
It seems to be the curl option `CURLOPT_UPLOAD` which is set in [guzzle 3 `CurlHandle:factory()`](https://github.com/guzzle/guzzle3/blob/master/src/Guzzle/Http/Curl/CurlHandle.php#L127). When I uncomment `$curlOptions[CURLOPT_UPLOAD] = true;` the request works.

aaa2000:
Using the PHPUnit annotation @runTestsInSeparateProcesses fixes also Guzzle5

justincy:
@adri When I comment out that line the request succeeds but the test fails because the POST body isn't set.

adri:
Yes indeed, I haven't found out the difference between the intercepted and the direct request yet.   I looked at the HttpClient what curl options are provided. Something there must set the body and make the request not "timeout" (error posted previously). Maybe it is related to the CURLOPT_READFUNCTION, I have to check.

justincy:
It looks like I may be using version 3 wrong. It's treating by post body string as a stream. It also expects the `Content-Length` header to contain the body size for `CURLOPT_INFILESIZE`. That behavior is undocumented so you'd think they would calculate that on their but it doesn't appear to be doing that.

justincy:
Though the direct request is still working, so maybe cURL is just really smart.

justincy:
I can get the requests to pass and almost all pass if I comment out line 28 of `HttpClient` in VCR:

```php
curl_setopt_array($ch, $request->getCurlOptions());
```

So Guzzle must be doing something unexpected with the curl opts. Perhaps it's related to streaming, since VCR intercepts streams and sends them all as strings via `CURLOPT_POSTFIELDS`.

justincy:
That's the `CURLOPT_UPLOAD` that you were talking about. Commenting that out works for me now. I think that wasn't working earlier for me because it was detecting form data and handling it differently and so the response from httpbin was different than expected. I'm testing with JSON now to avoid that. I think I'm close to fixing VCR for guzzle3.

justincy:
I think I've fixed them all.

To fix Guzzle 3 I had to flat out ignore `CURLOPT_UPLOAD`. That could be a bad idea.

To fix Guzzle 4 I first changed when the `CURLOPT_READFUNCTION, CURLOPT_INFILESIZE` assertion was being done so that `CURLOPT_INFILESIZE` didn't have to be set before `CURLOPT_READFUNCTION`. That didn't actually fix Guzzle 4 though. The problem there was that they were setting `CURLOPT_READFUNCTION` by default and then later setting `CURLOPT_POSTFIELDS` for small strings without bothering to set `CURLOPT_INFILESIZE`. So I changed it to ignore `CURLOPT_READFUNCTION` when the body has already been set via `CURLOPT_POSTFIELDS`.

Looks like the tests failed for Guzzle 4. I don't know why. They pass for me.

justincy:
Now we see green!

adri:
Great work! I'll review.

src/VCR/Util/CurlHelper.php
```diff
@@ -175,24 +175,39 @@ public static function setCurlOptionOnRequest(Request $request, $option, $value,
             case CURLOPT_HEADER:
             case CURLOPT_WRITEFUNCTION:
             case CURLOPT_HEADERFUNCTION:
+            case CURLOPT_UPLOAD:
```
adri: As far as I see [`CURLOPT_UPLOAD`](http://curl.haxx.se/libcurl/c/CURLOPT_UPLOAD.html) should be used when uploading a file in conjunction with a read function. So it should be still possible to upload files using the `@` notation when we ignore the option.

src/VCR/Util/CurlHelper.php
```diff
@@ -175,24 +175,39 @@ public static function setCurlOptionOnRequest(Request $request, $option, $value,
...
+     * @param resource $curlHandle cURL handle associated with the request.
+     */
+    public static function validateCurlPOSTBody(Request $request, $curlHandle = null)
+    {
+        $readFunction = $request->getCurlOption(CURLOPT_READFUNCTION);
+        if (is_null($readFunction)) {
+            return;
+        }
+        
+        // Guzzle 4 sometimes sets the post body in CURLOPT_POSTFIELDS even if
```
adri: When did you come across this case? As far as I see it's either an upload (with a read function) or a string. See   [CurlFactory](https://github.com/guzzle/guzzle/blob/ddbbba92efbf3d843baf467dd699b4647105c64a/src/Handler/CurlFactory.php#L224).

src/VCR/Util/CurlHelper.php
```diff
@@ -175,24 +175,39 @@ public static function setCurlOptionOnRequest(Request $request, $option, $value,
             case CURLOPT_HEADER:
             case CURLOPT_WRITEFUNCTION:
             case CURLOPT_HEADERFUNCTION:
+            case CURLOPT_UPLOAD:
                 // Ignore header, file and writer functions.
                 // These options are stored and will be handled later in handleOutput().
                 break;
-            case CURLOPT_READFUNCTION:
```
adri: Removing handling the curl option here was probably necessary because it is not clear if `CURLOPT_POSTFIELDS` have been set yet, right? As far as I understand this came from a special case when using Guzzle 4. Do you think it is possible to leave it here and check for the curl option because Guzzle 4 sets it before the `CURLOPT_READFUNCTION` option?

tests/integration/guzzle/3/ExampleHttpClient.php
```diff
@@ -24,6 +24,25 @@ public function get($url)
...
         return null;
     }
+    
+    public function post($url, $body)
+    {
+        $client = new Client();
+        
+        try {
+            $request = $client->post($url);
+            $request->setBody($body, 'application/json');
```
adri: Just for me to know, is the 'application/json' parameter changing anything how Guzzle sends the request? (Except from a `Content-Type` header?

tests/integration/guzzle/4/ExampleHttpClient.php
```diff
@@ -22,4 +22,20 @@ public function get($url)
 
         return null;
     }
+
+    public function post($url, $body)
+    {
+        $client = new Client();
+        
+        try {
+            $response = $client->post($url, ['body' => $body]);
```
adri: PHP-VCR is still supporting PHP 5.3, please use `array()`.

tests/integration/guzzle/5/ExampleHttpClient.php
```diff
@@ -22,4 +22,20 @@ public function get($url)
 
         return null;
     }
+
+    public function post($url, $body)
+    {
+        $client = new Client();
+        
+        try {
+            $response = $client->post($url, ['body' => $body]);
```
adri: PHP-VCR is still supporting PHP 5.3, please use `array()`.

tests/integration/guzzle/test/VCR/Example/ExampleHttpClientTest.php
```diff
@@ -20,19 +24,18 @@ class ExampleHttpClientTest extends \PHPUnit_Framework_TestCase
 
     public function setUp()
     {
-        // Configure virtual filesystem.
-        vfsStream::setup('testDir');
-        \VCR\VCR::configure()->setCassettePath(vfsStream::url('testDir'));
+        \VCR\VCR::configure()->setCassettePath('/tmp');
```
adri: We should still use vfsStream. In theory this shouldn't be a problem to use. Writing to the file system is not handy for repeatable tests since the fixture file then needs to be deleted.

src/VCR/Util/CurlHelper.php
justincy: You linked to the current version which is Guzzle 5. The case I was handling occurs in [Guzzle 4](https://github.com/guzzle/guzzle/blob/4.2.3/src/Adapter/Curl/CurlFactory.php). In `getDefaultOptions` you can see that `CURLOPT_READFUNCTION` is set by default while in `applyBody` you can have `CURLOPT_POSTFIELDS` set instead of `CURLOPT_UPLOAD` and `CURLOPT_INFILESIZE`.

src/VCR/Util/CurlHelper.php
justincy: I don't think we can move it. As you see in Guzzle 4 that I just linked to, `CURLOPT_READFUNCTION` is set before `CURLOPT_POSTFIELDS`. Plus this also avoids the problem of needing to set `CURLOPT_INFILESIZE` before `CURLOPT_READFUNCTION`.

tests/integration/guzzle/3/ExampleHttpClient.php
justincy: No it doesn't change anything other than setting the `Content-Type` header.

justincy:
@adri I stopped using vfsStream because it fails for some reason when recording POST requests. I put it back to show the errors. How do you suggest we handle this?

aaa2000:
There are issues that can not be fixed in vfsStream https://github.com/mikey179/vfsStream/wiki/Known-Issues

> If you use an object with a destructor that opens a vfsStream URL, vfsStream can become unavailable before the open occurs, causing it to fail with a WARNING. E.g. in PHPUnit, make sure to manually destroy the object in the test or in tearDown().

I think the problem is that VCR use a singleton, the same instance of videorecorder and same cassette is used during all the tests.

So, I see two way:
- the @runTestsInSeparateProcesses annotation on top of the class https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.runTestsInSeparateProcesses
```
/**
 * @runTestsInSeparateProcesses
 * @preserveGlobalState disabled
 */
```
- or not using \VCR\VCR class in ExampleHttpClientTest and bootstrap.php

```
   public function setUp()
    {
        // Configure virtual filesystem.
        vfsStream::setup('testDir');
        $this->configuration = new \VCR\Configuration();
        $this->configuration->setCassettePath(vfsStream::url('testDir'));
        $this->httpClient = new \VCR\Util\HttpClient();
        $this->factory = \VCR\VCRFactory::getInstance($this->configuration);
        $this->videorecorder = new \VCR\Videorecorder($this->configuration, $this->httpClient, $this->factory);
    }
...
    protected function requestGETIntercepted()
    {
        $this->videorecorder->turnOn();
        $this->videorecorder->insertCassette('test-cassette.yml');
        $info = $this->requestGET();
        $this->videorecorder->turnOff();

        return $info;
    }

    protected function requestPOSTIntercepted()
    {
        $this->videorecorder->turnOn();
        $this->videorecorder->insertCassette('test-cassette.yml');
        $info = $this->requestPOST();
        $this->videorecorder->turnOff();

        return $info;
    }
```

adri:
Great analysis @aaa2000 thank you. 

When [turning off VCR](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Videorecorder.php#L135) the cassette instance is actually unset. But you are right the storage will then be retrieved from the Factory which is the same instance for the same file.

There might be another way. That assertion for a writable file leads to quite some problems, see for example #103. Maybe on some systems `is_writeable` returns `false` although writing would be possible.

I propose to remove the assertion. Directly after asserting for writeablility the file is written to. So when it's not writeable that code will just fail anyway. That means that in [Yaml](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Storage/Yaml.php#L48) and in [Json](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Storage/Json.php#L20) we could remove the assertion and fix the tests. @justincy What do you think, want to give it a try? My local test got green then. I guess then you can confirm, we can merge.

justincy:
@adri Fixed.

adri:
Awesome, well done.

PR #128, opened 2015-07-02
Add support for HTTP Continue headers

stevemiketa:
Requests that return an HTTP 100 Continue header before returning the remainder of the response are currently not processed properly in `src/VCR/Util/HttpUtil.php`. A typical response header looks like:
```
HTTP/1.1 200 OK
Date: Thu, 02 Jul 2015 22:32:22 GMT
Server: ***
...

<content>
```

When a continue header is encountered, the response header looks like: 
```
HTTP/1.1 100 Continue

HTTP/1.1 200 OK
Date: Thu, 02 Jul 2015 22:32:22 GMT
Server: ***
...

<content>
```
This PR adds support for parsing that additional Continue header.

Thoughts / Comments ?

### DIFF
--- a/src/VCR/Util/HttpUtil.php
+++ b/src/VCR/Util/HttpUtil.php
@@ -61,7 +61,12 @@ class HttpUtil
      */
     public static function parseResponse($response)
     {
-        list($rawHeader, $rawBody) = explode("\r\n\r\n", $response, 2);
+        if (strpos($response, 'HTTP/1.1 100 Continue') === 0) {
+            list($continueHeader, $rawHeader, $rawBody) = explode("\r\n\r\n", $response, 3);
+        }
+        else {
+            list($rawHeader, $rawBody) = explode("\r\n\r\n", $response, 2);
+        }
 
         // Parse headers and status.
         $headers = self::parseRawHeader($rawHeader);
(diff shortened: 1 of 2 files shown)


adri:
Thank you @stevemiketa! 

I wonder if there might be multiple continue headers arriving. Do you know? It seams like, reading the following on stack overflow:
http://stackoverflow.com/questions/11359276/php-curl-exec-returns-both-http-1-1-100-continue-and-http-1-1-200-ok-separated-b

stevemiketa:
@adri No problem!  Based on what I've read [here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html) and [here](https://support.urbanairship.com/entries/59909909--Expect-100-Continue-Issues-and-Risks), it would seem that there would only be one continue header and a final header, since Continue is designed to be able to reject a request based on headers alone.  Once the headers are accepted, the body sends and you get the final status code.

The above solution would work with a smaller code change, but would nix RFC allowed functionality.

stevemiketa:
Here's the code solution for this issue, should @adri not want to merge this PR. 

```php
\VCR\VCR::getEventDispatcher()->addListener(VCREvents::VCR_BEFORE_HTTP_REQUEST,
    function (Event $event) {
        $event->getRequest()->setHeader('Expect', '');
    });
```

stevemiketa:
Hey @adri, found [this](http://greenbytes.de/tech/webdav/rfc2616.html#rfc.section.10.1) which looks like you were right. It has to support multiple 100 Continue headers. I will update

adri:
Thanks a lot @stevemiketa, it's merged.

PR #133, opened 2015-07-28
Added SOAPAction header

SuRaMoN:
http://www.w3.org/TR/2000/NOTE-SOAP-20000508/#_Toc478383528


adri:
Thanks a lot for your contribution @SuRaMoN.

This should only be added when the SOAP version is 1.1. SOAP 1.2 uses the already implemented `ContentType: [...] action=` header.

The test `testShouldInterceptSoapLibrary` fails because of this.

laland:
#141 solves this issue, so you can close this one

Issue #108, opened 2015-02-23
url_stat should handle a second param $flags


dbu:
Stream::url_stat has a second parameter $flags that would tell whether we want to resolve symlinks or not. we should probably handle this parameter in StreamProcessor. No idea what the right thing to do would be, unfortunately.

I just know that when we have a symlinked folder, is_link still returns false, making e.g. the symfony Filesystem::remove try to delete a folder instead of removing the symlink.

We tried some stuff in #106 but that was a series of misunderstandings. No idea what should be done for proper behaviour.

dbu:
probably we should call [lstat](http://php.net/lstat) instead when STREAM_URL_STAT_LINK is set.

not sure if we really want to respect STREAM_URL_STAT_QUIET to supress errors or trigger_error on errors.

adri:
I stole this class (as mentioned somewhere in the docs) from here:
https://github.com/antecedent/patchwork/blob/master/src/Preprocessor/Stream.php

It seems to be an issue and maybe this will help:
https://github.com/antecedent/patchwork/commit/4310769f47e45abcb0bb621360024f7c4126fe62

dbu:
hm, i think @antecedent will have a problem with symlinks in his project. @antecedent did you ever check if is_link still works when the patchwork pre processor is loaded?

we should port that change you link and then add the lstat case. will see if i get to it.

PR #137, opened 2015-08-19
Honor silent flag & prevent error handler catch

mdarse:
This is a more generic fix to https://github.com/php-vcr/php-vcr/pull/35, it should fix https://github.com/php-vcr/php-vcr/issues/108 too.

### DIFF
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -280,9 +280,10 @@ class StreamProcessor
      *
      * @link http://www.php.net/manual/en/streamwrapper.url-stat.php
      *
-     * @param  string $path The file path or URL to stat.
+     * @param  string  $path  The file path or URL to stat.
+     * @param  integer $flags Holds additional flags set by the streams API.
      *
-     * @return integer      Should return as many elements as stat() does.
+     * @return integer        Should return as many elements as stat() does.
      */
     public function url_stat($path, $flags)
     {
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -284,18 +284,15 @@ class StreamProcessor
      *
      * @return integer      Should return as many elements as stat() does.
      */
-    public function url_stat($path)
+    public function url_stat($path, $flags)
     {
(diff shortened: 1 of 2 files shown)


src/VCR/Util/StreamProcessor.php
```diff
@@ -280,22 +280,23 @@ public function stream_tell()
...
     {
         $this->restore();
-        try {
+        if ($flags & STREAM_URL_STAT_QUIET) {
+            set_error_handler(function() {
+                // Use native error handler
+                return false;
+            });
             $result = @stat($path);
-        } catch (\ErrorException $e) {
```
adri: Looks nice thank you. I wonder though, if we don't catch the exception PHPUnit might throw ErrorExceptions when running in `processIsolation`. Did you test this by any chance?

mcrumm:
:+1: for this.  I'm using [kahlan](https://github.com/crysalead/kahlan) for specs, and ran into #33 when I added a dependency that referenced an existing namespace. This PR corrects that problem.

In an attempt to answer your question about PHPUnit and process isolation, I ran the following:

```bash
$ vendor/bin/phpunit --process-isolation tests/VCR/Util/StreamProcessorTest.php
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /Users/mike/Code/php-vcr/php-vcr/phpunit.xml

..........

Time: 5.93 seconds, Memory: 4.25Mb
```

Hopefully that's enough to green-light this patch.  If there's more I can do to help verify this as working, please let me know.

mcrumm:
D'oh, but of course PHPUnit didn't gripe about it.  PHP-VCR's dependencies don't share namespaces.  I'll try a more specific test and report back.

mcrumm:
Well, I had to write a PHPUnit test for my project to be sure, but I can confirm that this works under process isolation for at least PHPUnit v3.7.38 (VCR's version from `composer.lock`) and PHPUnit v5.1.3.

adri:
Alright, let's merge then.

mcrumm:
Thanks!

mdarse:
Thanks for merging this :+1:

PR #144, opened 2015-10-13
provide more information when not in recording mode

dbu:
the Videorecorder should report more information to figure out what request was not found. the not-recording mode is usually on a ci server, so any debugging information will be helpful.


adri:
Pardon for the delay and thanks for this :)

PR #149, opened 2016-01-25
also build lowest version of dependencies on travis to see if composer.json is correct

dbu:


.travis.yml
```diff
@@ -24,15 +24,17 @@ matrix:
...
         php: 5.6
+      - env: TEST_DIR=. COMPOSER_FLAGS="--prefer-lowest"
+        php: 5.6
   allow_failures:
     - php: hhvm-nightly
     - php: 7.0
 
 before_install:
-  - composer install && cd ${TEST_DIR}
+  - composer self-update && cd ${TEST_DIR}
```
dbu: no need to install twice, right?

.travis.yml
```diff
@@ -24,15 +24,17 @@ matrix:
...
 
 install:
-  - composer install
+  - composer update $COMPOSER_FLAGS
```
dbu: --prefer-lowest only works with update

.gitignore
```diff
@@ -4,3 +4,4 @@
 coverage/*
 notes.md
 tests/fixtures/cassette*
+composer.lock
```
dbu: https://getcomposer.org/doc/02-libraries.md#lock-file - i would go further than that and say its harmful as it hides possible problems with the things specified in composer.json. a project using this library will never use your lock file so it could end up with different versions than the ones from the lock file.

dbu:
looks like i used up your github allocation. either try to set it up with shared keys or just restart the build in an hour or so.

adri:
Thanks! 

Seems like Phpunit gets downgraded from 3.7.38 to 3.7.18 in:
PHP 5.5 https://travis-ci.org/php-vcr/php-vcr/jobs/104633570 and
PHP 5.6 https://travis-ci.org/php-vcr/php-vcr/jobs/104633570

Could we try to change the minimum phpunit dependency to 3.7.38, but less than 4?

dbu:
updated the lowest version requirement for phpunit, lets see

.gitignore
renatomefidf: The advantage for composer.lock in my opinion is regarding the library contributors, it creates a safe setup for them start coding on it, what do you think?

.travis.yml
renatomefidf: For sure :)

.travis.yml
renatomefidf: We could also have another one which has the composer install then? Which would be regarding the idea I said above to keep the lock for the library contributors. What do you think?

.gitignore
dbu: i have to respectfully but strongly disagree ;-)

being a library, the library lock file is not used by the users of php-vcr. the only thing tracking the lock file can do here is hiding problems. if this was an application, the assumption that the lock file helps would be correct. but in a reusable library, any user might at any point update dependencies. if composer.json does not limit dependencies to working versions, things will fail for them.

renatomefi: I totally understand your opinion, I don't know if I was clear enough, I know the users can just update everything and it won't do anything, I mean in a moment someone clones our project to contribute to php-vcr core, the lock file is useful to have a safe state to start patching for us. In that manner do you think it's useful?

dbu: i am more worried about the experience of the users of this library than the experience of contributors. even for contributors, things will work fine without the .lock file unless something is horribly broken, in which case we would be glad to know its broken.

but well, anyways. i am not the maintainer of this library, if you are still convinced its a good idea, i can remove this line. i would however really not test on travis-ci with composer install, but keep doing composer update to not have the CI lie about whether things still work.

dbu: i think the .lock file adds more harm than doing benefit, but its your call really.

i updated this PR to not touch the ignore / .lock file

dbu:
i had a problem with the git repo i used, so had to create a new pull request. the things that are relevant are in #168

PR #150, opened 2016-01-25
add branch alias for master branch

dbu:
this would help people to depend on `1.2.*@dev` instead of `dev-master` when needing not yet released features.


adri:
Thanks again @dbu! :)

PR #154, opened 2016-03-07
Fixes #153 - SoapClient::__getLastRequest() returns NULL even with tr…

### LINKED ISSUE #153: SoapClient::__getLastRequest() returns NULL even with trace option set to 1
We have a project that integrates with a SOAP API, for which we use generated code that uses a class that derives from SoapClient. We are using VCR for our test suite.

When running the tests, calling __getLastRequest() on the SoapClient derived class returns NULL when VCR is on, even with trace option set to 1 in the SoapClient constructor options. This happens only with some PHP versions. It happens with PHP version 5.4.14, but not with version 5.5.9.

By debugging the test that fails due to this issue, I can see that the protected $request field is correctly set to the SOAP request in [VCR\Util\SoapClient::__doRequest()](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/Util/SoapClient.php#L50), but is changed to NULL just after stepping out of the method and while still inside of __soapCall().

Looking at do_soap_call() in [ext/soap/soap.c](https://github.com/php/php-src/blob/PHP-5.4.14/ext/soap/soap.c#L2710), there is an xmlFreeDoc(request) after do_request(), which makes me suspect the request is freed differently depending on the PHP version, but I can not investigate further.

There is a quick fix for this, which is to save a copy of the original SOAP request in __doRequest() instead of the original request, by concatenating it with an empty string.

buduchail:


castarco:
Hi @buduchail , could you provide a more detailed description of your pull request please?

Thank you for your time!

renatomefidf:
@buduchail I've checked the mentioned bug, thanks for making this!
Could you add the version 5.4.14 specifically to travis?
Also do you know other ways to fix this? Maybe casting to string or cloning the result?

If you also could provide us with the php bug report would be nice to check the best way to fix this. Thank you for your time!

Issue #132, opened 2015-07-27
Do identical requests get replayed in sequence?


bensheldon:
Will PHP-VCR record multiple requests to the same matching resource and then play them back in order? This is implemented in the [Ruby VCR Library](https://relishapp.com/vcr/vcr/v/2-9-3/docs/request-matching/identical-requests-are-replayed-in-sequence) but it seems like it isn't implemented in PHP-VCR (unless I'm just using the library wrong).

Sorry to ask a question, but I couldn't find the answer in the docs or quickly in the code. Thanks for the help.

adri:
Hey Ben. Thank a lot for your question. This feature is indeed currently not implemented.

I'm interested... how did you come across this usecase?

bensheldon:
@adri well nuts :-)

For the usecase, we're writing behat stories. Many of our stories involve some sort of flow like:

1. read out and verify the original value (GET)
2. modify the value (PUT)
3. refetch the value in step 1(GET) and verify that it's been modified

So we're making the same GET request twice, but expecting a different response in order to verify that the PUT is operating correctly (as well as we're properly flushing caches within our own application).

bensheldon:
@adri Also, we (@pantheon-systems), would be willing to sponsor/bounty this feature. Please email me at ben@getpantheon.com if you'd be interested.

EdwardIII:
As @bensheldon as a quick workaround when we do this we just pop in a separate cassette for step 3.

chrisrichard:
#160 should enable this scenario

esbenp:
Ping @bensheldon @adri what do you guys think of this fix https://github.com/esbenp/php-vcr/commit/278a11b75621cff16d04e212eb9258a735efc844 ?

If we have a test that has the following requests:
* (1) `GET /resource`
* (2) `PUT /resource`
* (3) `GET /resource`

This commit keeps track of the order of occurrence of requests, meaning (1) is the 1st occurrence of (GET /resource) and (3) is the 2nd. When the test is re-run, when running (3) PHP-VCR will now look for a saved request which is the 2nd (GET /resource).

Right now the tests are failing but if you like the fix I can start looking into the tests and make an actual PR. Let me know what you think.

bensheldon:
@esbenp yes, thank you. That looks like exactly what I'm looking for. Thank you for resurrecting this issue. 🐣

esbenp:
Perfect @bensheldon. What do you say @adri, should I fix tests and make a real PR?

[...]

PR #161, opened 2016-07-13
Enable recording of identical requests

esbenp:
This PR will introduce the tracking of identical requests, thus enabling PHP-VCR to record different responses of the same request within the same cassette. This closes issue #132 

Right now PHP-VCR only tracks one occurrence of a given request
per cassette. An example of why this is troublesome is given below.
- (1) Request a resource from an external endpoint (GET /resource)
- (2) Request an update to that resource (PUT /resource)
- (3) Re-request the resource to confirm it was persisted remotely (GET /resource)

Requests (1) and (3) are identical, however the responses are expected
to be different. PHP-VCR will only record (1) and use the saved version
on (3).

This commit keeps track of the order of occurrence of requests, meaning
(1) is the 1st occurrence of (GET /resource) and (3) is the 2nd. When
the test is re-run, when running (3) PHP-VCR will now look for a
saved request which is the 2nd (GET /resource).

Ref #132

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-curl": "*",
         "beberlei/assert": "2.*",
         "symfony/yaml": "2.*|3.*",
-        "symfony/event-dispatcher": "~2.1|3.*"
+        "symfony/event-dispatcher": "^2.4|^3.0|^4.0"
     },
 
     "require-dev": {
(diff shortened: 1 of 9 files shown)


castarco:
For me this is a +1, It would be great if someone else reviews this merge request to ensure that we won't break anything (I'm a maintainer but I'm relatively new to this codebase).

esbenp:
Okay, I am neither completely fluent in the codebase, however it seems that @adri has gone AWOL

fmasa:
What is current status of this PR, could I help somehow?

PR #166, opened 2016-10-17
Improve versions constraints to better compatibility

renatomefi:
When merging it, please do with squash since it has loads of try outs!


adri:
Also see #149 for reference

PR #173, opened 2016-10-29
php-cs-fixer introduced

nstapelbroek:
Hi!

I've added a codestyle checker and fixed some codestyle issues in the src code. 
Please let me know you you'll need anything changed. 

Some notes:
- I did not touch the test directory, but I'm happy to do so if requested. 
- Since fixing the method name violations in `src/VCR/Util/StreamProcessor.php` and `src/VCR/LibraryHooks/StreamWrapperHook.php` would break the API, I've added an exclude to the rules instead. I would be happy to change this, but I'm not sure if its possible or if you simply don't want to because of API.

### DIFF
--- a/src/VCR/Videorecorder.php
+++ b/src/VCR/Videorecorder.php
@@ -204,7 +204,8 @@ class Videorecorder
      *
      * @return Response                Response for the intercepted request.
      * @throws \BadMethodCallException If there was no cassette inserted.
-     * @throws \LogicException         If the mode is set to none or once and the cassette did not have a matching response.
+     * @throws \LogicException         If the mode is set to none or once and
+     *                                 the cassette did not have a matching response.
      */
     public function handleRequest(Request $request)
     {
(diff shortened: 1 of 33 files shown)


src/VCR/Videorecorder.php
```diff
@@ -228,13 +229,20 @@ public function handleRequest(Request $request)
             return $response;
         }
 
-        if (VCR::MODE_NONE === $this->config->getMode() || VCR::MODE_ONCE === $this->config->getMode() && $this->cassette->isNew() === false) {
+        if (VCR::MODE_NONE === $this->config->getMode()
+            || VCR::MODE_ONCE === $this->config->getMode()
+            && $this->cassette->isNew() === false
+        ) {
             throw new \LogicException(
-                "The request does not match a previously recorded request and the 'mode' is set to '{$this->config->getMode()}'. "
```
nstapelbroek: Not sure how this `{$this->config->getMode()}` works. I've removed it when adding sprintf :see_no_evil:

renatomefi:
@nstapelbroek thanks for your contribution!

Can you put a phpcs file configuration? Also make the test available at composer test, this way you don't have to pass the flags while calling it on travis.

Also when you put the config file let's see which kind of rules do we want and how they would look like in the project!

renatomefi:
Oh, and I'm comfortable having only the php-cs-fixer in the repository. So no need for phpcs. You can add the `--dry-run`  and `--diff` arguments to the command when running the fixer. If it returns some changes the result code would be 1, failing the build. 👍

nstapelbroek:
Updated the code, let me know if you need any additional adjustments :smile:

tests/VCR/Storage/YamlTest.php
```diff
@@ -62,7 +62,7 @@ public function testIterateFirstNestedObject()
     public function testIterateSecondNestedObject()
     {
         $this->iterateAndTest(
-            "-". "\n"
+            "-" . "\n"
```
renatomefi: can you put a rule for double and single quotes?

renatomefi: use the `single_quote` rule

renatomefi: canso you also check if there is any change if you enable `braces`? I feel like some old if statements without braces are not being fixed, might be an impression tho!

nstapelbroek:
I've added the requested rules. Personally I think the single quote rule on the php-cs-fixer creates more inconsistency then it prevents. My preference would be to refactor this in a separate issue. 

If you agree on the changes, you can merge anyway :smile:

renatomefi:
@nstapelbroek why do you think it generates more inconsistency?

nstapelbroek:
@renatomefi i think the suggestion was good. Only the php-cs-fixer made things more inconsistent because it wont touch strings with a variable. Honestly, we should also take care of the double quoted lines before merging this. What do you think?

renatomefi:
@nstapelbroek but it's ok if the double quotes actually have variables, what do you think?

renatomefi:
As we discussed I think it's ok to have double quotes to the special strings! If you want you can do the same for https://github.com/php-vcr/phpunit-testlistener-vcr

Thank you very much!

nstapelbroek:
Thank you for taking the time to review!

Awesome. Will probably work on the testlistener-vcr this weekend.

PR #175, opened 2016-11-15
Set POST method when CURLOPT_POSTFIELDS is set

webspectyler:
According to the curl documentation, using CURLOPT_POSTFIELDS implies CURLOPT_POST. This patch ensures that when CURLOPT_POST is not set and CURLOPT_POSTFIELDS is set, the request method is POST.

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -160,6 +160,7 @@ class CurlHelper
                 } else {
                     $request->setBody($value);
                 }
+                $request->setMethod('POST');
                 break;
             case CURLOPT_HTTPHEADER:
                 foreach ($value as $header) {
--- a/tests/VCR/Util/CurlHelperTest.php
+++ b/tests/VCR/Util/CurlHelperTest.php
@@ -127,6 +127,16 @@ class CurlHelperTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($expected, $request->getHeaders());
     }
 
+    public function testSetCurlOptionOnRequestPostFieldsSetsPostMethod()
+    {
+        $request = new Request('GET', 'example.com');
+        $payload = json_encode(array('some' => 'test'));
+
+        CurlHelper::setCurlOptionOnRequest($request, CURLOPT_POSTFIELDS, $payload);
+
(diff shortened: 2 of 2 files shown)


adri:
That's awesome @webspectyler! Thank you for your contribution.

May I ask you to add a test to the [CurlHelperTest](https://github.com/php-vcr/php-vcr/blob/master/tests/VCR/Util/CurlHelperTest.php) to ensure this functionality will be verified automatically? Thanks a lot!

adri:
This is awesome. The integration test for SOAP is failing but this is not on you.

Thank you for your work. Will merge it after resolving the integration test.

renatomefi:
Integration tests are fixed now! Thanks

TeslaDethray:
This is a breaking change. It should warrant a 2.0.0.

adri:
Thanks for the heads up @TeslaDethray! Is this blocking you from updating? 

Do you by any chance have a code example that we can use to make a unit test?

TeslaDethray:
@adri Yes, I've had to pin at 1.3.1 because my the API my app hits uses PUT requests.

The unit test included in this PR actually tests the error pretty nicely - the 'GET' is being overridden just as other methods would be. Change 'GET' to 'PUT' in the extant test:
```bash
    public function testSetCurlOptionOnRequestDoesNotRuinMethods()
    {
        $method = 'ANYTHING BUT POST';
        $domain = 'example.com';
        $request = new Request($method, $domain);
        $headers = ["Host: $domain",];

        CurlHelper::setCurlOptionOnRequest($request, CURLOPT_HTTPHEADER, $headers);

        $this->assertEquals(['Host' => $domain,], $request->getHeaders());
        $this->assertNotEquals($request->getMethod(), $method);
    }
```
The array version of `$args` that `CurlHook::__callStatic` is getting from Guzzle includes an element `[CURLOPT_POSTFIELDS] => 'PUT'`.

webspectyler:
I'm just expecting behavior as I understand it from the cURL documentation:
https://curl.haxx.se/libcurl/c/CURLOPT_POSTFIELDS.html

If I'm misinterpreting "Using CURLOPT_POSTFIELDS implies CURLOPT_POST", please disregard this issue.

PR #177, opened 2016-11-27
Fixes #103 is_writable on NFS mounts

### LINKED ISSUE #103: Error when checking for casette file is_writable in vagrant

adri:
### Context

In #103 @dbu and @Danno040 discovered that there is an issue when using vagrant/virtualbox + NFS  mounts and trying to access.

### What has been done

- Overwrite `is_writable` calls with our own implementation that makes sure to restore stream wrappers for that call and intercept after again
- Added test a test

### How to test

- Checkout the source code of this project, switch to the `fix/is_writable_nfs` branch
- `composer install`
- Create a `Vagrantfile` with this content

    ```
    Vagrant.configure(2) do |config|
      config.vm.synced_folder ".", "/vagrant", type: "nfs"
      config.vm.network "private_network", type: "dhcp"
      config.vm.box = "laravel/homestead-7"
    end
    ```
- Run `vagrant up` and `vagrant ssh`
- Comment the  `appendCodeTransformer` line in the constructor of `StreamProcessor`
- `cd /vagrant && vendor/bin/phpunit --filter testWritableForUnknownUser`
- The test should fail
- Uncomment that line
- The test should pass

### Notes

- This is done to make sure to get [this `if` in the php-source](https://github.com/php/php-src/blob/2104bea5d756dfa40b605a4a2765a3bc4637a76c/ext/standard/filestat.c#L777) code to evaluate to `true`. Doing so will use the default behaviour which hits the C access functions instead of a manual access check based on the `uid`, `gid` and `mode`.

### DIFF
--- a/src/VCR/CodeTransform/StreamProcessorCodeTransform.php
+++ b/src/VCR/CodeTransform/StreamProcessorCodeTransform.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace VCR\CodeTransform;
+
+class StreamProcessorCodeTransform extends AbstractCodeTransform
+{
+    const NAME = 'vcr_stream_processor';
+
+    private static $patterns = array(
+        '/(?<!::|->|\w_)\\\?is_writable\s*\(/i' => '\VCR\Util\StreamProcessor::is_writable(',
+        '/(?<!::|->|\w_)\\\?is_writeable\s*\(/i' => '\VCR\Util\StreamProcessor::is_writable(',
+    );
+
+    /**
+     * @inheritdoc
+     */
+    protected function transformCode($code)
+    {
+        return preg_replace(array_keys(self::$patterns), array_values(self::$patterns), $code);
+    }
+}
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -2,6 +2,7 @@
 
 namespace VCR\Util;
 
+use VCR\CodeTransform\StreamProcessorCodeTransform;
 use VCR\Configuration;
 use VCR\CodeTransform\AbstractCodeTransform;
 
@@ -61,6 +62,10 @@ class StreamProcessor
     {
         if ($configuration) {
             static::$configuration = $configuration;
+
+            $codeTransformer = new StreamProcessorCodeTransform();
+            $codeTransformer->register();
+            $this->appendCodeTransformer($codeTransformer);
         }
     }
 
@@ -573,6 +578,24 @@ class StreamProcessor
                 break;
         }
         $this->intercept();
+
+        return $result;
+    }
+
+    /**
+     * Tells whether the filename is writable.
+     *
+     * @link http://php.net/manual/en/function.is-writable.php
+     * @param string $filename The filename being checked.
+     */
+    public static function is_writable($filename)
+    {
+        $hook = new self;
+        $hook->restore();
+
+        $result = is_writable($filename);
+
+        $hook->intercept();
 
         return $result;
     }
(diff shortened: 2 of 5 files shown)


adri:
@renatomefi The test fails because of `'http://www.webservicex.net/` is down ;) 
If we only had php-vcr to be less dependent on external services haha. Any idea which other service to use?

renatomefi:
@adri omg, is this a project sin? Choosing web services which are gonna fail! haha

Let's wait a bit, but if it stays like this we have to think about something else :(

renatomefi:
Builds passes now :)

renatomefi:
@adri the test specified works, but when I run the whole suit I get another issue:
```
PHPUnit 5.6.2 by Sebastian Bergmann and contributors.

...............................................................  63 / 267 ( 23%)
............................................................... 126 / 267 ( 47%)
............................................................... 189 / 267 ( 70%)
...............................................E............... 252 / 267 ( 94%)
...............                                                 267 / 267 (100%)

Time: 736 ms, Memory: 8.00MB

There was 1 error:

1) VCR\Util\StreamProcessorTest::testStreamMetadata
chown(): Operation not permitted

/vagrant/src/VCR/Util/StreamProcessor.php:565
/vagrant/src/VCR/Util/StreamProcessor.php:565
/vagrant/tests/VCR/Util/StreamProcessorTest.php:150
```

Now commenting the line you said:
```
vagrant@homestead:/vagrant$ vendor/bin/phpunit
PHPUnit 5.6.2 by Sebastian Bergmann and contributors.

...............................................................  63 / 267 ( 23%)
............................................................... 126 / 267 ( 47%)
............................................................... 189 / 267 ( 70%)
........................................F......E............... 252 / 267 ( 94%)
...............                                                 267 / 267 (100%)

Time: 682 ms, Memory: 8.00MB

There was 1 error:

1) VCR\Util\StreamProcessorTest::testStreamMetadata
chown(): Operation not permitted

/vagrant/src/VCR/Util/StreamProcessor.php:565
/vagrant/src/VCR/Util/StreamProcessor.php:565
/vagrant/tests/VCR/Util/StreamProcessorTest.php:150

--

There was 1 failure:

1) VCR\Util\StreamProcessorTest::testWritableForUnknownUser
tests/fixtures/streamprocessor_data_wrong_user is not writable while intercepting.
Failed asserting that false is true.

/vagrant/tests/VCR/Util/StreamProcessorTest.php:70

ERRORS!
Tests: 267, Assertions: 367, Errors: 1, Failures: 1.

```

adri:
Cleaning up

PR #179, opened 2016-12-02
Add function to easily retrieve request matcher names.

mprelude:
### Context

PHP-VCR expects me to enable matchers by name, but provides no way to get a list of names of available matchers from which to select.

### What has been done

- Add a function to get all names for available matchers.

### How to test

- Use the function.

### DIFF
--- a/src/VCR/Configuration.php
+++ b/src/VCR/Configuration.php
@@ -266,6 +266,16 @@ class Configuration
     }
 
     /**
+     * Returns a list of valid names for request matchers, to pass to enableRequestMatchers.
+     *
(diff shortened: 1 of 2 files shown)


adri:
Great idea! Thanks. Could you please add a unit test in https://github.com/php-vcr/php-vcr/blob/master/tests/VCR/ConfigurationTest.php#L31? 
The reason for this is that without having tests on all new functionality we can end up in accidentally breaking things when refactoring. The test ensures that your functionality will stay :)

mprelude:
@adri I've added a unit test, hope this is sufficient. :)

PR #181, opened 2016-12-03
Override the content type and encoding in the XML to work similar to SoapClient

jacques:
### Context

PHP-VCR ignores the encoding required from the WSDL which one needs to manually pass in.

### What has been done

- If encoding is in options in the SoapHook set the content type header for the encoding
- In the XML if the encoding is in the options replace UTF-8 with the given encoding.

### How to test

- Test SoapClient against http://ws.dev.freepaid.co.za/airtimeplus/

### Todo

- [ ]

### Notes

-
-


renatomefi:
@jacques thanks for you PR!

Can you write some tests where the XML isn't utf-8?

Thank you agian!

PR #183, opened 2016-12-09
Fix permission deny warrning if the fixture file has read-only permission

filevgenij:
If the fixture file has read-only permission function fopen throw a warning because r+ mode is used


renatomefi:
thanks @filevgenij 

Even though the building is still running can you provide a test where the file is non writable and see if the behavior is still maintained?

Thank you agian!

JeroenVanOort:
As this has not been followed up, I think it is safe to say that the suggested behaviour is not been subject to popular demand. Therefore, I'm closing this PR.

PR #186, opened 2016-12-30
Always override request method if CURLOPT_CUSTOMREQUEST is set

tmdk:
### Context
The cURL option CURLOPT_CUSTOMREQUEST can be used to send PUT/DELETE requests with message bodies. Currently, `CurlHelper` handles this option by calling `Request::setMethod`. This is not sufficient, because the handling of other options (such as CURLOPT_POSTFIELDS) also call `setMethod` and therefore override the custom request method. Seeing that there already is a way to set cURL options on the Request object, I've created a patch that makes php-vcr behave more like cURL itself.

### What has been done

- `Request::getMethod` returns the value of CURLOPT_CUSTOMREQUEST if set.
- `CurlHelper::setCurlOptionOnRequest` sets CURLOPT_CUSTOMREQUEST on the request instead of calling `setMethod`

### How to test

- See added test cases
- Try sending a PUT/DELETE request with a message body (I'm using Guzzle 6.2.1)

### Notes

- Did not run SOAP Weather test cases as the service is broken for me atm.

### DIFF
--- a/src/VCR/Request.php
+++ b/src/VCR/Request.php
@@ -148,6 +148,10 @@ class Request
      */
     public function getMethod()
     {
+        if ($this->getCurlOption(CURLOPT_CUSTOMREQUEST) !== null) {
+            return $this->getCurlOption(CURLOPT_CUSTOMREQUEST);
+        }
+
         return $this->method;
     }
 
--- a/tests/VCR/RequestTest.php
+++ b/tests/VCR/RequestTest.php
@@ -215,4 +215,19 @@ class RequestTest extends \PHPUnit_Framework_TestCase
         $request = new Request('GET', 'http://example.com:5000/foo?param=key');
         $this->assertEquals('example.com:5000', $request->getHost());
     }
+
+    public function testCurlCustomRequestOverridesMethod()
+    {
+        $postRequest = new Request('POST', 'http://example.com');
+        $getRequest = new Request('GET', 'http://example.com');
+
+        $this->assertEquals('POST', $postRequest->getMethod());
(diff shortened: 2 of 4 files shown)


src/VCR/Request.php
```diff
@@ -148,7 +148,11 @@ public function getBody()
      */
     public function getMethod()
     {
-        return $this->method;
+        if ($this->getCurlOption(CURLOPT_CUSTOMREQUEST) !== null) {
+            return $this->getCurlOption(CURLOPT_CUSTOMREQUEST);
+        } else {
+            return $this->method;
```
renatomefi: Maybe you can remove this from the else statement

tests/VCR/RequestTest.php
```diff
@@ -215,4 +215,13 @@ public function testGetHostReturnsBothHostAndPort()
         $request = new Request('GET', 'http://example.com:5000/foo?param=key');
         $this->assertEquals('example.com:5000', $request->getHost());
     }
+
+    public function testCurlCustomRequestOverridesMethod()
```
renatomefi: Maybe you can make also a POST test in order to prove compatibility

tmdk:
@renatomefi thanks for looking at the PR. I've added some more assertions, please let me know if you still think tests are missing!

senechaux:
@renatomefi Did you have the opportunity of checking the new changes made by @tmdk ?

tmdk:
@renatomefi ping! :)

renatomefi:
@tmdk has been a long time, I'm sorry!

Could you please rebase the PR?

I think the code looks good now and I'm planning to merge and tag it!

PR #189, opened 2017-02-02
composer: bump to PHP 7.0+ as only developed version

TomasVotruba:
See: http://php.net/supported-versions.php

uu

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     ],
 
     "require": {
+        "php": "^7:0",
         "ext-curl": "*",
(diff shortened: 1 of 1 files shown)


renatomefi:
Thanks, for now we are going to keep "support" for the older versions.
Possibly in the next major we are going to change it, for now the current major is not adding complex features, it's more about keeping things working.

Thanks for keeping an eye on us and have a good weekend!

TomasVotruba:
Bumping version doesn't remove support for older versions

Issue #136, opened 2015-08-18
Response header status failing Braintree library checks due to status being saved as a string


acorncom:
Howdy, I'm using the Braintree PHP libraries and was happily getting setup with your library (much appreciated).

I've run into a snag though (that may keep me from using it for now unless I fork your repo).  The issue seems to be as follows (this is the JSON cassette for my test):

```json
[{
    "request": {
        "method": "POST",
        "url": "https:\/\/api.sandbox.braintreegateway.com\/merchants\/<merchant id snipped>\/subscriptions",
        "headers": {
            "Host": "api.sandbox.braintreegateway.com",
            "Accept": "application\/xml",
            "Content-Type": "application\/xml",
            "User-Agent": "Braintree PHP Library 3.2.0",
            "X-ApiVersion": "4"
        },
        "body": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<subscription>\n <payment-method-token>test-nonce<\/payment-method-token>\n <plan-id type=\"integer\">2<\/plan-id>\n<\/subscription>\n"
    },
    "response": {
        "status": {
            "http_version": "1.1",
            "code": "422",
        },
        "more":"snipped"
```

My issue is that the Braintree library does this test:

```php
        if($responseCode === 200 || $responseCode === 201 || $responseCode === 422 || $responseCode == 400) {
            return Braintree_Xml::buildArrayFromXml($response['body']);
        } else {
            Braintree_Util::throwStatusCodeException($responseCode);
```

Due to the use of the `===`, I'm seeing an exception thrown when using the cassette (the status code is a string when retrieving values from the cassette) but not when retrieving directly from the API (where the status code is an integer as specified in the RFC on status codes: http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html)

It looks like that's a bug in your library.  I'm happy to send in a pull request if that'd be helpful, but a larger fundamental question is: **do you want to fix this?"  Seems like it would be good to, but I imagine it could be a backwards-compatibility breaking change with unintended consequences for other folks using your library ...

Thoughts?

acorncom:
Further, potentially relevant information:
- PHP 5.6.10
- Storage systems tried for the library: YAML and JSON (both exhibit the same behavior)

adri:
Hey @acorncom, thanks a lot for your time to investigate in this issue. Of course it should be fixed. 

[...]

PR #192, opened 2017-03-10
Fixes CurlHook to return integer HTTP status codes

louismrose:
### Context

As discussed in #136, the Braintree SDK uses strict comparison (===) against the HTTP status code obtained via cURL. This test currently fails when using PHP VCR canned responses, because PHP VCR stores and returns the HTTP status code as a string. 

### What has been done

- Adjusted CurlHook to cast HTTP status codes to `int` before returning them. This matches PHP curl's behaviour, which seems to always returns an `int` for `CURLINFO_HTTP_CODE`.
- Considered but ultimately decided not to change `VCR\Response` to return status codes as integers: although this fixed the problem, it felt like changing the type of data that is returned from a public method could break other people's code.

### How to test

- Run the new unit test `CurlHookTest::testShouldReturnCurlInfoStatusCodeAsInteger`
- Run the new version of PHP VCR code with Braintree SDK and observe that Braintree no longer throws a status code exception.

### What needs to be done

- 75ed534 should be reverted if this PR is merged.

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -103,7 +103,7 @@ class CurlHelper
                 }
                 break;
             case CURLINFO_HTTP_CODE:
-                $info = $response->getStatusCode();
+                $info = (int)$response->getStatusCode();
                 break;
             case CURLINFO_SIZE_DOWNLOAD:
                 $info = $response->getHeader('Content-Length');
--- a/tests/VCR/LibraryHooks/CurlHookTest.php
+++ b/tests/VCR/LibraryHooks/CurlHookTest.php
@@ -176,8 +176,29 @@ class CurlHookTest extends \PHPUnit_Framework_TestCase
         $infoHttpCode = curl_getinfo($curlHandle, CURLINFO_HTTP_CODE);
         curl_close($curlHandle);
 
-        $this->assertEquals(200, $infoHttpCode, 'HTTP status not set.');
+        $this->assertSame(200, $infoHttpCode, 'HTTP status not set.');
+
         $this->curlHook->disable();
+    }
+
+   /**
+    * @see https://github.com/php-vcr/php-vcr/issues/136
+    */
+    public function testShouldReturnCurlInfoStatusCodeAsInteger()
+    {
+      $stringStatusCode = "200";
+      $integerStatusCode = 200;
+      $this->curlHook->enable($this->getTestCallback($stringStatusCode));
+
+      $curlHandle = curl_init('http://example.com');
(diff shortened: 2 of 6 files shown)


louismrose:
It looks like the build was failing because the WSDL file used in tests is no longer available:

http://wsf.cdyne.com/WeatherWS/Weather.asmx?WSDL

I've fixed this as well by using the Internet Archive's most recent working URL:

https://web.archive.org/web/20161213164919/http://wsf.cdyne.com/WeatherWS/Weather.asmx?WSDL

louismrose:
It seems that using Internet Archive's mirror wasn't reliable, so I've brought the WSDL into the Git repo. It can now be served by Github to the CI environment:

https://raw.githubusercontent.com/louismrose/php-vcr/issue136-integer-status-codes-for-braintree/tests/fixtures/weather.wsdl

Note that 75ed534 should be reverted if this PR is merged.

https://raw.githubusercontent.com/php-vcr/php-vcr/master/tests/fixtures/weather.wsdl

louismrose:
Hi there, @adri @castarco. Is there anything I could do to help along a review of this PR?

I believe that it fixes a bug (which prevents PHP-VCR from being used with Braintree #136), and also has fixed an issue with builds on the CI server (which might be causing other PRs to fail the build).

renatomefi:
Hello @louismrose 
Thanks for the effort on it and sorry for the delay!
I'm afraid to know what happens when someone had a strict comparison using the strings in their code, this means we might be breaking BC, what do you think about it?
I'll also check with adri what he thinks about it!

louismrose:
Thanks for the feedback @renatomefi! You make a good point about the potential for breaking BC with strict comparisons.

I wonder how often this might arise though, because `curl_getinfo(..., CURLINFO_HTTP_CODE)` [returns an integer](http://php.net/manual/en/function.curl-getinfo.php#example-4788). So perhaps it would be unusual for a strict comparison against a string returned from PHP VCR, because that code would break when PHP VCR was turned off and PHP's normal curl functions were active?

renatomefi:
@louismrose sure, for the normal user it shouldn't be a problem, I was afraid of user with custom matchers or wrappers, but it should be a really small number.

Thanks for fixing the tests, do you mind splitting it in another PR? Only if it's not too much work for you, the reason being is if we have an issue with any of them reverting is easier, and I can also in the other PR see where I can host the wsdl!

renatomefi:
@louismrose we've fixed the tests in master, could you please rebase your PR?
I'm considering putting this feature on 1.4.0

renatomefi:
@louismrose I've cherry-picked your fix and put it on #220 
Thanks so much for your contribution!

PR #193, opened 2017-03-20
Fix stream emulation

markstory:
### Context

Stream API emulation is incomplete, because `headers` cannot be accessed out of the wrapper data so client code that works with native streams + curlwrappers fails when emulated by VCR.

### What has been done

- I've added the missing behavior to access the `headers` key which is populated by the native `http` and `https` stream implementations when PHP is compiled with curlwrappers support (the default on osx)
- I've had to skip all the SOAP related tests. The WSDL file they are based on is throwing IIS errors. See http://wsf.cdyne.com/WeatherWS/Weather.asmx?WSDL

### How to test

- Use the CakePHP or Aura HTTP clients and turn on VCR. Without these changes you get mostly cryptic errors.

### Todo

N/A

### Notes

N/A


markstory:
Closing as #230 incorporates these changes with test fixes.

PR #198, opened 2017-05-26
Add support for CURLINFO_HEADER_OUT

alnorth:
### Context

When `curl_setopt($ch, CURLINFO_HEADER_OUT, true)` has been called before a request is sent you should then be able to call `curl_getinfo($ch, CURLINFO_HEADER_OUT)` after the request to get a description of the request that's been sent. Prior to this patch calling `curl_getinfo($ch, CURLINFO_HEADER_OUT)` with VCR enabled would actually cause an exception and break the tests entirely.

### What has been done

Implementing this in VCR proved to be a little tricky, as the Response objects don't know anything about the Request objects. I've added a little helper function and called it in CurlHook::curlExec so that the request info is passed on from Request to Response. I'm not 100% satisfied that this is the right way to go, but it's the least disruptive way I could think of of doing it.

This also required a function to format the Request as a string. I've mimicked curl's output for a simple GET request, but doing it for a POST with a body would take much more work. It's got enough there fulfil my purposes at the moment. It feels like the sort of thing that there may be a standard PHP function out there to do, but I haven't found one.

In summary, this does feel like a slightly hacky way of going about this, but I'm not sure how to do better.

### How to test

I've included tests at multiple levels. Hopefully enough to show that this works. Just run the phpunit tests as normal.

### DIFF
--- a/tests/VCR/ConfigurationTest.php
+++ b/tests/VCR/ConfigurationTest.php
@@ -136,9 +136,9 @@ class ConfigurationTest extends \PHPUnit_Framework_TestCase
     public function testGetStorage()
     {
         $class = $this->config->getStorage();
-        $this->assertTrue(in_array('Iterator', class_implements($class)));
-        $this->assertTrue(in_array('Traversable', class_implements($class)));
-        $this->assertTrue(in_array('VCR\Storage\AbstractStorage', class_parents($class)));
+        $this->assertContains('Iterator', class_implements($class));
+        $this->assertContains('Traversable', class_implements($class));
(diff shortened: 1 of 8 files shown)


alnorth:
BTW, I'm just about to head off on holiday for a week, so I won't be able to make any tweaks immediately.

alnorth:
Anything I can do to help out further? The unit test failures only seem to be related to the SOAP issues mentioned in #192.

renatomefi:
Hello @alnorth sorry for the huge delay!

Could you rebase your PR, the tests should pass now!
After that I'll take a closer look into the code, but I'll first merge the PRs with bug fixes for a new patch!

Thank you

Issue #200, opened 2017-06-07
Always base64 encode body


slootjes:
When trying this cool library I have the issue that images with various headers are not detected as binary. This causes json or yaml storage to break and thus the library can not do it's job. When doing some testing I found that simply always base64 encoding the body solves my issue and all my recordings (including the ones with binary data) work flawlessly.

So either having the option to use a custom storage or to simply skip the binary checks and always encode the body as base64 would solve the issue for me. Any opinions?

higidi:
Closing as stalled.

The current storage implementations already base64-encode bodies when binary content is detected (via `Content-Transfer-Encoding: binary` or `application/x-gzip`). An "always base64" toggle would bloat all cassettes including plain-text responses with no benefit.

For image or other binary responses not detected correctly today, the better fix is to improve the binary-detection heuristic. If you have a specific content-type that isn't being caught, please open a new issue with the response headers and body type — happy to look at extending the detection logic.

PR #203, opened 2017-06-20
Assume binary response by default

slootjes:
### Context
I found out that responses with images were not saved because currently it is assumed that the body can be encoded as string (see issue #200 ). Because of how headers vary much I decided to assume binary by default and add exceptions for text responses instead.

### What has been done
Assume binary by default and based on headers text is assumed.

### How to test
I added tests for this function in "testIsBinaryResponse"


slootjes:
Need any more help with this?

PR #160, opened 2016-06-11
Added 'strict' mode which requires all requests on cassette to play in order

chrisrichard:
Feedback appreciated.

PR #204, opened 2017-07-07
Strict mode

hectorj:
Same as https://github.com/php-vcr/php-vcr/pull/160 but with the conflicts resolved


hectorj:
Test aren't passing :/

PR #206, opened 2017-07-25
Make handling of empty strings in CURLOPT_POSTFIELDS consistent.

alnorth:
### Context

Before this change empty strings in CURLOPT_POSTFIELDS weren't treated consistently. CurlHelper called `setBody` with an empty string, whereas `Request::fromArray` wouldn't call `setBody` at all, leaving the body as `null`. This means that stored requests are never matched because `null === ''` is false.

### What has been done

My aim was to make the handling consistent. I've done this by making CurlHelper ignore empty strings, so that it doesn't call `setBody` at all.

### How to test

- The testSetCurlOptionOnRequestPostFieldsEmptyString test has been added to CurlHelperTest.

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -157,7 +157,9 @@ class CurlHelper
                     if (count($value) == 0) {
                         $request->removeHeader('Content-Type');
                     }
-                } else {
+                } elseif (!empty($value)) {
+                    // Empty values are ignored to be consistent with how requests are read out of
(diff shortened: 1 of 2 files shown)


alnorth:
As far as I can tell the only unit test issues are with the SOAP handler, which is that same with a lot of the current pull requests.

renatomefi:
@alnorth looks good, I'll put this for 1.4 since user making strict comparisons can have some BC issues!

Could you rebase also this PR?

Thank you again

renatomefi:
@alnorth thanks for your contribution :+1:

PR #207, opened 2017-07-28
Fix issue if multiple headers are returned

GM-Alex:
### Context

If multiple headers are returned the and `CURLOPT_HEADER` is set to `false` the first header will be removed but the response contains all other headers.

### What has been done

- Removes all headers expect the last one.

### How to test

- The behavior can be discovered by a curl request which returns multiple headers.

### DIFF
--- a/src/VCR/Util/HttpUtil.php
+++ b/src/VCR/Util/HttpUtil.php
@@ -66,8 +66,10 @@ class HttpUtil
     public static function parseResponse($response)
     {
         $response = str_replace("HTTP/1.1 100 Continue\r\n\r\n", '', $response);
-            
-        list($rawHeader, $rawBody) = explode("\r\n\r\n", $response, 2);
+
+        $responseChunks = explode("\r\n\r\n", $response);
+        $rawBody = array_pop($responseChunks);
+        $rawHeader = array_pop($responseChunks);
 
         // Parse headers and status.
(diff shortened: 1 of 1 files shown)


renatomefi:
Hello @GM-Alex 

Thanks for contributing! Could you please rebase your PR and also provide some tests for this issue?

renatomefi (changes requested):
Hello @GM-Alex

Thanks for contributing! Could you please rebase your PR and also provide some tests for this issue?

higidi:
Closing this as superseded.

The multi-header-block handling in `HttpUtil::parseResponse()` was reworked after this PR was opened. The "leading header blocks leak into the body" problem is now handled generically instead of by popping the last two `\r\n\r\n` chunks:

- 100 Continue header blocks — #128, released in 1.2.6
- Generalized leading header blocks (proxy CONNECT / informational responses), rewriting `parseResponse` into a skip-loop — #467, released in 1.11.1
- Guard for responses without a body separator — #476 (issue #272), released in 1.11.2
- 100 Continue with an extra header, regression coverage — #480 (issue #335), on master for the next release

The `array_pop`-the-last-two-chunks approach here would also truncate any body that itself contains a blank line (`\r\n\r\n`), which the current parser explicitly preserves. Please reopen if you can still reproduce on 1.11.2+ with a concrete response.

Thanks for the report and the patch.

Issue #112, opened 2015-03-02
Curl getinfo not stored


dbu:
curl getinfo is not recorded. when replaying, i end up with "Not implemented: CURLINFO_EFFECTIVE_URL (1048577)" - maybe we should have an option to activate recording this information. Not sure if it should be on by default, its quite heavy and often not relevant.

JeroenVanOort:
I'm closing this issue because it has not seen any activity for many years. If anyone feels this is still a problem, please open a new issue.

PR #210, opened 2017-08-29
Add recording of curl infos

endorama:
Hello, first of all thanks for this project!

### Context
Trying to test a third party library, I found out that PHP-VCR is not storing curl informations, and this was breaking the tests.

The issue #112 confirmed the missing feature.

### What has been done

- added recording of curl informations
- tests

### How to test

- run tests of `tests/VCR/ResponseTest.php`, should be all green
- register the call produced by this piece of code:

  ```php
  $curl = curl_init('http://example/');
  curl_setopt($curl, CURLOPT_HTTPGET, true);
  $body = curl_exec($curl);
  $contentType = curl_getinfo($curl, CURLINFO_CONTENT_TYPE);
  curl_close($curl);
  ```
- the playback should work as expected

### Notes

Following concerns from @dbu on #112 about making this configurable: making this option configurable is a big tasks; passing around the configuration object require to touch the code in lots of places. For clarity on the main business logic around this PR I didn't do it. Please let me know if it's a requirement.

Thank you

### DIFF
--- a/tests/VCR/ResponseTest.php
+++ b/tests/VCR/ResponseTest.php
@@ -120,6 +120,15 @@ class ResponseTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($curlOptions, $response->getCurlInfo());
     }
 
+    public function testRestoreCurlInfoFromArray()
+    {
+        $expectedCurlOptions = array('option' => 'value');
+        $response = new Response(200, array(), null, $expectedCurlOptions);
+        $restoredResponse = Response::fromArray($response->toArray());
+
+        $this->assertEquals($expectedCurlOptions, $response->getCurlInfo());
+    }
+
     public function testToArray()
     {
         $expectedArray = array(
@@ -131,7 +140,10 @@ class ResponseTest extends \PHPUnit_Framework_TestCase
             'headers'   => array(
                 'host' => 'example.com'
             ),
-            'body'      => 'Test response'
+            'body'      => 'Test response',
+            'curl_info' => [
+                'content_type' => 'text/html'
+            ]
         );
 
         $response = Response::fromArray($expectedArray);
--- a/tests/VCR/ResponseTest.php
+++ b/tests/VCR/ResponseTest.php
@@ -141,9 +141,9 @@ class ResponseTest extends \PHPUnit_Framework_TestCase
                 'host' => 'example.com'
             ),
             'body'      => 'Test response',
-            'curl_info' => [
+            'curl_info' => array(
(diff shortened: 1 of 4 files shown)


dbu:
looks sane to me. the build error looks like network problems of travis-ci, so maybe @adri can just restart the failed builds to see if thats all.

will this break with existing recordings? if it does, the code should probably check if the infos are stored and otherwise fall back to the previous behaviour.

endorama:
@dbu If you have a look at [this line](https://github.com/php-vcr/php-vcr/pull/210/files#diff-a8b7f8b0221612d697367b6301b38fe8R99): 

```php
public static function fromArray(array $response) {
    [...]
    isset($response['curl_info']) ? $response['curl_info'] : array()
    [...]
}
```

If the curl_info is not present in the stored data, is initialized as an empty array. This should keep backward compatibility. 

From my understanding this change should not affect any current recording.

Please let me know if I'm wrong!

dbu:
looks plausible to me. so :+1: from me!

Spea:
I had the same issue when using the [solarium](https://github.com/solariumphp/solarium) client, would love to see this PR merged :)

renatomefi:
@endorama thanks for the PR!

The tests are fixed, would you please rebase your PR?

Thank you again

endorama:
@renatomefi I've rebased and the tests are red. I'll fix them as soon as possible!

tests/VCR/ResponseTest.php
```diff
@@ -131,7 +140,10 @@ public function testToArray()
             'headers'   => array(
                 'host' => 'example.com'
             ),
-            'body'      => 'Test response'
+            'body'      => 'Test response',
+            'curl_info' => [
```
renatomefi: Short array syntax isn't supported in PHP 5.3

dbu:
this looks great! tests will continue to work with data recorded before this change, right? then yes, go for it and always record it. maybe add an entry to the CHANGELOG file?

renatomefi:
@dbu indeed they do! I only modified one test to prove it the new data works but the other ones are still without curl_info stored on it!

CHANGELOG has been moved to the releases page, I can make an special note there!

endorama:
@renatomefi thank you for this project :) glad to have contributed!

dantleech:
Would this PR be merged into master?

dantleech:
Would this PR be merged to master?

JeroenVanOort:
This has been merged as part of https://github.com/php-vcr/php-vcr/pull/259

PR #212, opened 2017-09-18
Add a file check to StreamProcessor

mjanser:
If a file which StreamProcessor should process doesn't exist, `fopen`
gives a warning. In PHPUnit context this warning will be translated into
an exception which will then leave StreamProcessor in disabled state.

### DIFF
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -171,6 +171,10 @@ class StreamProcessor
      */
     public function stream_open($path, $mode, $options, &$openedPath)
     {
+        if ('r' === substr($mode, 0, 1) && !is_file($path)) {
+            return false;
+        }
+
         $this->restore();
 
         if (isset($this->context)) {
--- a/tests/VCR/Util/StreamProcessorTest.php
+++ b/tests/VCR/Util/StreamProcessorTest.php
@@ -58,7 +58,19 @@ class StreamProcessorTest extends \PHPUnit_Framework_TestCase
         );
     }
 
-    public function testStreamOpenShouldNotFailOnInexistingFile()
+    public function streamOpenFileModesWhichDoNotCreateFiles()
+    {
+        return array(
+            array('r'),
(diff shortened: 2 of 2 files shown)


renatomefi:
@mjanser thanks for checking it and contributing!

Could you please provide a test with a failing file? Also please rebase your PR

Thank you

mjanser:
Added a test and also changed the condition a bit to do the check only when the mode doesn't create the file.

src/VCR/Util/StreamProcessor.php
```diff
@@ -171,6 +171,12 @@ protected function shouldProcess($uri)
      */
     public function stream_open($path, $mode, $options, &$openedPath)
     {
+        $subMode = substr($mode, 0, 1);
+
+        if (in_array($subMode, array('r', 'w', 'a')) && !is_file($path)) {
```
mjanser: I'm not sure this is correct now because according to the [documentation of fopen()](http://php.net/manual/en/function.fopen.php) all modes will create the file if it doesn't exist except `r` and `r+`. But I'm not an expert on this topic :)

renatomefi: Oh, actually the "a" is an error, sorry. r and w don't create files

renatomefi: To verify that you can add the other modes on the dataProvider in the test and see the file is created

mjanser: Ah, ok, but for `w` it also says "If the file does not exist, attempt to create it". Is the documentation wrong then?

renatomefi:
Indeed. I missed that part in the docs. I'll fix it shortly

src/VCR/Util/StreamProcessor.php
renatomefi: Indeed, I lost this last sentence in the docs, and I even tried to double check every option :/

renatomefi:
@mjanser thanks a lot for your help and contributions!
Have a great week

Issue #213, opened 2017-10-04
Host Header gets overwritten whens setting the URL


jimktrains:
If a host header is set, either from a cassette or by hand, and then the URL is set (as is done in the constructor), the `setURL` method will overwrite the `Host` header.  This behavior is ill-advised and can cause issues when the `Host` header from curl contains a port, or if they were made to be different for a specific reason.

higidi:
Closing — fixed on master.

`Request::setUrl()` now only sets the `Host` header when none is already present (`false === $this->hasHeader('Host')`). The overwrite described here no longer occurs.

If you still see the `Host` header being clobbered on a current release, please reopen with a repro.

PR #215, opened 2017-10-04
Issue #213. Don't set Host header in setUrl if already set.

jimktrains:
### Context

If a header has been set manually, do not overwrite it with the one from the URL.
Issue #213 

### What has been done

- Do not overwrite the Host Header in setUrl if it is already set
- Test to verify

### How to test

- vendor/bin/phpunit tests/VCR/RequestTest.php

### DIFF
--- a/src/VCR/Request.php
+++ b/src/VCR/Request.php
@@ -132,7 +132,9 @@ class Request
     public function setUrl($url)
     {
         $this->url = $url;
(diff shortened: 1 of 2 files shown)


renatomefi:
Hello @jimktrains ,

It looks good, thanks for fixing it, could you please rebase your PR with the current master?

PR #221, opened 2017-10-19
Ugly hack for NTLM authentication

kostajh:
### Context

php-vcr doesn't work when using Curl (via Guzzle) to interact with a SOAP service that uses Microsoft NTLM authentication.

The problem is that the response headers are always 401 HTTP 1.1/Unauthorized but the body of the. request contains HTTP 200 OK response, along with other headers.

### What has been done

- An ugly hack that gets NTLM working for my particular use case.

### How to test

- Find an NTLM service and set up phpro/soap-client to use Guzzle to interact with it.

### Todo

- [ ] Find a more elegant way to support NTLM

### Notes

n/a


JeroenVanOort:
I think it would be wiser to find out what changes PHP-VCR makes to trigger this weird condition than to use hacks like this to work around the symptoms.

PR #222, opened 2017-10-21
Apply suggested php-cs-fixer patches

nstapelbroek:
### Context
I'm starting to work on updating php-cs-fixer to version 2.0 but first I'd like to create some sort of clean state. Not sure if the code-style violations were intentional.

### What has been done

- Applied the php-cs-fixer violation diffs which fixed the indenting and single quoted strings. 

### How to test

- If the build passes, we should be good right :+1: ?


renatomefi:
Thank you :+1:

Issue #224, opened 2017-10-23
PHPCsFixer approach


renatomefi:
# Context
- php cs fixer is now a required dev library, even though it's only used being build or development it causes trouble because of php 5.3 version.

# The idea
- Let's remove it from composer and make a `composer cs-fix` script which downloads it and run it the same way it currently runs
- Change travis to do the same but only on php 7.1

nstapelbroek:
I'll give this a shot!

PR #225, opened 2017-10-27
Feature/remove legacy php cs fixer

nstapelbroek:
### What has been done
- Tried implementing #224
- Instead of writing a small script and committing it in the repo, i thought I could just write an if statement that downloads the binary in the PHP 7.1 build target. You current setup is fail-save towards php-cs-fixer issues anyway.


.php_cs.dist
```diff
@@ -1,46 +1,23 @@
...
-        ->setFinder(
-            PhpCsFixer\Finder::create()
-                ->exclude('vendor')
-                ->exclude('docs')
-                ->in(__DIR__)
-        )
-        ->setCacheFile($cacheFile)
-        ->setUsingCache(true)
-        ->setRiskyAllowed(true);
+$cacheFile = getenv('TRAVIS') ? getenv('HOME') . '/.php_cs.cache' : __DIR__ . '/.php_cs.cache';
```
renatomefi: maybe we can drop this?

.php_cs.dist
```diff
@@ -1,46 +1,23 @@
...
+$config = PhpCsFixer\Config::create()
+    ->setRules(array(
```
renatomefi: since we're running it on php 5.6+ it's safe to use the short array syntax

.php_cs.dist
```diff
@@ -1,46 +1,23 @@
...
+        'single_quote' => true,
+        'braces' => true,
+    ))
+    ->setFinder(
+        PhpCsFixer\Finder::create()
+            ->exclude('vendor')
+            ->exclude('docs')
+            ->in(__DIR__)
+    )
+    ->setCacheFile($cacheFile)
```
renatomefi: can be removed as well?

.php_cs.dist
```diff
@@ -1,46 +1,23 @@
...
-            PhpCsFixer\Finder::create()
-                ->exclude('vendor')
-                ->exclude('docs')
-                ->in(__DIR__)
-        )
-        ->setCacheFile($cacheFile)
-        ->setUsingCache(true)
-        ->setRiskyAllowed(true);
+$cacheFile = getenv('TRAVIS') ? getenv('HOME') . '/.php_cs.cache' : __DIR__ . '/.php_cs.cache';
+$config = PhpCsFixer\Config::create()
```
renatomefi: you can do early return here, no need for the variable

.travis.yml
```diff
@@ -38,15 +38,14 @@ before_install:
 
 install:
   - composer update --prefer-source -o $COMPOSER_FLAGS
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar -O $(pwd)/php-cs-fixer; fi
```
renatomefi: Please remove the pwd, shouldn't be necessary since we aren't changing the working directory

composer.json
```diff
@@ -27,8 +27,7 @@
         "phpunit/phpunit": "^4.8|^5.0",
         "sebastian/version": "^1.0.3|^2.0",
         "mikey179/vfsStream": "^1.2",
-        "lapistano/proxy-object": "dev-master#d7184a479f502d5a0f96d0bae73566dbb498da8f",
-        "friendsofphp/php-cs-fixer": "^2.7|^1.12"
```
renatomefi: :tada:

.travis.yml
```diff
@@ -38,15 +38,13 @@ before_install:
 
 install:
   - composer update --prefer-source -o $COMPOSER_FLAGS
+  - if [ "${PHP_CS_FIXER}" = "true" ]; then wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar && php ./php-cs-fixer-v2.phar fix --dry-run --diff $TRAVIS_BUILD_DIR; fi
```
renatomefi: Why not put it on script section after phpunit?

renatomefi:
Amazing, thank you @nstapelbroek

PR #226, opened 2017-11-02
Setting CURLOPT_NOBODY automatically sets method to HEAD

snelg:
### Context
Fixes problem when using php-vcr with Guzzle 3.

Per PHP docs, setting curl option `CURLOPT_NOBODY` to true automatically sets the method to HEAD
http://php.net/manual/en/function.curl-setopt.php
Guzzle 3 properly sets the request method to HEAD, but VCR's curl hook defaults back to GET.

### What has been done

- Added a clause to the switch statement in `CurlHelper::setCurlOptionOnRequest`

### How to test

Demonstration of issue pre-patch:

- ```composer require guzzlehttp/guzzle:~3.0```
- 
```php
VCR::turnOn();
VCR::insertCassette('head.yml');
$client = new \Guzzle\Http\Client();
$request = $client->head('http://headers.jsontest.com');
$response = $request->send();
VCR::turnOff();
```
- ```cat head.yml```
The request method in head.yml is "GET" instead of "HEAD"

### DIFF
--- a/src/VCR/Util/Assertion.php
+++ b/src/VCR/Util/Assertion.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace VCR\Util;
 
 use Assert\Assertion as BaseAssertion;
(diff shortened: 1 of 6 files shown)


higidi:
Thanks for reporting this and for the fix, @snelg — appreciated. This branch has diverged quite a bit from current master (paths and signatures have changed since 2017), so I've picked up the underlying fix in #484, adapting your CURLOPT_NOBODY case and test to the current codebase. Closing this in favor of that PR, but the core idea here is yours.

PR #228, opened 2017-11-03
Strict mode (fixed #160 and #204)

snelg:
Tiny style tweaks to #160 and #204 so that Travis builds all pass.

Issue #159, opened 2016-06-10
Why doesn't VCR throw an exception if requests on cassette are NOT played back


chrisrichard:
Shouldn't it throw an exception if the component under test does not make the expected HTTP requests?

chrisrichard:
I'm going to create a PR that allows eject to take a parameter to control this behavior. If specified, eject will throw an exception the the cassette wasn't played back in its entirety.

Is there a reason why this is a bad idea?

chrisrichard:
#160

higidi:
Keeping open as a feature request.

PR #276 is the most complete existing implementation of strict mode — it throws when a cassette isn't fully replayed. A maintainer decision on that PR would be the fastest path to landing this feature.

PR #229, opened 2017-11-03
Strict mode allows recording new cassettes

snelg:
### Context
This is an addition to #228
I think that "strict" mode should behave like "once" mode when the requested cassette does not exist.
The original "strict" implementation has no way to record initial cassettes.
When I need a new cassette (e.g. I create a new TestCase), I have to change my code to ```setMode('once')```, create the cassette, and then change my code back to ```setMode('strict')```.
I'm submitting this as a separate pull request, because #228 was supposed to be only a style-cleaned version of the original #160. This is a functional change from that PR.

### What has been done

- If cassette is new, then "strict" mode behaves like "once" mode

### How to test

- ```setMode('strict')```
- Insert a non-existent cassette
- Trigger a request (curl, stream wrapper, etc)
- Fail!

### Todo

- N/A

### Notes

- N/A


higidi:
Closing as stalled — superseded by PR #276.

This was an add-on to PR #228 (strict mode records new cassettes in 'once' fashion). PR #276 is the canonical strict-mode implementation and the right place to continue this discussion. The feature request is tracked in issue #159.

PR #231, opened 2017-11-09
Fix RuntimeException with SplFileInfo->isFile()

snelg:
### Context
With stream_wrapper handling enabled, SplFileInfo->isFile() triggers a RuntimeException for non-existing files.

### What has been done

- Simply added a try/catch around the line triggering the exception, ```$result = @stat($path);```

### How to test

- ```VCR::turnOn()```
- ```$path = new \SplFileInfo('missingFile');```
- ```$path->isFile()```

### DIFF
--- a/src/VCR/Util/Assertion.php
+++ b/src/VCR/Util/Assertion.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace VCR\Util;
 
 use Assert\Assertion as BaseAssertion;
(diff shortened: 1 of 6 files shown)


snelg:
Just noticed #137. I'm not sure if this undoes part of what that PR was trying to fix. This one is catching ```RuntimeException``` instead of ```ErrorException``` which was removed in the previous PR, so at least it's different in that regard. And the test I added for this PR fails without the try/catch.

PR #232, opened 2017-11-12
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06) and [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
 
     "require-dev": {
-        "phpunit/phpunit": "^4.8.35|^5.7",
+        "phpunit/phpunit": "^4.8.35|^5.7|^6.0",
         "sebastian/version": "^1.0.3|^2.0",
         "mikey179/vfsStream": "^1.2",
         "lapistano/proxy-object": "dev-master#d7184a479f502d5a0f96d0bae73566dbb498da8f"
--- a/tests/VCR/CassetteTest.php
+++ b/tests/VCR/CassetteTest.php
@@ -22,9 +22,12 @@ class CassetteTest extends TestCase
         $this->cassette = new Cassette('test', new Configuration(), new Storage\Yaml(vfsStream::url('test/'), 'json_test'));
     }
 
+    /**
+     * @expectedException VCR\VCRException
+     * @expectedExceptionMessage Cassette name must be a string, array given.
+     */
     public function testInvalidCassetteName()
     {
-        $this->setExpectedException('\VCR\VCRException', 'Cassette name must be a string, array given.');
         new Cassette(array(), new Configuration(), new Storage\Yaml(vfsStream::url('test/'), 'json_test'));
     }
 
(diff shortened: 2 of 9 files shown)


renatomefi:
Hello @gabriel-caruso.

Thanks for your contribution! I've been thinking about doing it for a while but we'll have to wait for version 2.0 to break this BC!

gabriel-caruso:
@renatomefi Support this new `namespace` isn't a BC. In version 2, support `PHPUnit 6` will be, and then we can work on that.

renatomefi:
But people with phpunit 4.0 won't be able to update it anymore, right?

gabriel-caruso:
Ow, right, now I get your point. So, let's wait for `version 2.0`. If you want, create a development branch, and I can drop `PHPUnit 4` support, and add version 6 :smile:

renatomefi:
Even though the 4.8.35 looks like a good solution as you did! Give me a moment to think about that!

renatomefi:
Thanks @carusogabriel 

I'll put it to v1.5 then!

Would you like to add also phpunit 6 to the list?

carusogabriel:
Gonna work on that!

carusogabriel:
I've added `PHPUnit 6` support. Changes:
- `setExpectedException` isn't available in version 6, so I substitute to `@expectedException` and `@expectedExceptionMessage` notations.
- In `VCR\Util\SoapClientTest::testDoRequestExpectingException` I check if this methods exists, to support variable expectation across all `PHPUnit` versions.

renatomefi:
@carusogabriel Nice solution to keep compatibility! 👍 
Do you think 6.5 is really the minimal version? We could try 6.0 or so to see what happens!
Did you have time to check the failed builds?

I really appreciate the time and effort you're putting here, it's really valuable! Sorry for asking you so much, if you don't have time to finish it I'll try to find some to do it!
Thanks again and have a great week!

carusogabriel:
@renatomefi No problem, I'm on vacations on college, so I'm spending my time here on GitHub :sweat_smile: 

I'm going to check if we can work with `PHPUnit ^6.0` and fix the failed tests.

carusogabriel:
@renatomefi I’m having trouble aliasing some `PHPUnit` classes. May you help me? I can’t figure out what I did wrong 😔

PR #236, opened 2017-12-08
Also test against PHP 7.2

arubacao:
### Context
PHP `7.2` has been released. 

### What has been done

- Updated the travis.yml file

### How to test

- Let travis do the job

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,10 @@ matrix:
       php: 7.1
     - env: TEST_DIR=. COMPOSER_FLAGS="--prefer-lowest"
       php: 7.1
+    - env: TEST_DIR=.
+      php: 7.2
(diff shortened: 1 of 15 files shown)


renatomefi:
Hey @tiagobrito thanks for the contribution!
It's not really necessary to test the lower reqs for 7.2, if we find the lowest compatible version of each library using 7.2 we are gonna limit it too much and not gain a lot from it! I do believe people using 7.2 has updated libraries.

We can introduce the lowest 7.2 during the next minor/major!

tiagobrito:
@arubacao You should rebase this with master, the php cs fixer is already working there.
Thanks for your contribution.

arubacao:
@renatomefi @tiagobrito done :)

renatomefi:
Thanks so much @arubacao

PR #238, opened 2017-12-12
Rewrite Travis CI using Build Stages

### LINKED ISSUE #233: Rewrite Travis configuration using Build Stages
The current Travis strategy is a bit messy if we check the change of directories and how the code style is checked, new feature Build Stages could help on that:
https://docs.travis-ci.com/user/build-stages/

Make sure the SOAP tests are different stages

carusogabriel:
### Context
This resolves #233 

### What has been done

- Rewrite Travis CI using [Build Stages](https://docs.travis-ci.com/user/build-stages).

### Todo
- [x] Simplify code
- [ ] Add PHP 7.2
- [ ] ~~Create stage for `Composer lowest`~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,18 @@
 language: php
 dist: precise
+sudo: false
 
-matrix:
-  fast_finish: true
-  include:
-    - env: TEST_DIR=.
-      php: 5.3
-    - env: TEST_DIR=. COMPOSER_FLAGS="--prefer-lowest"
-      php: 5.3
-    - env: TEST_DIR=.
-      php: 5.4
-    - env: TEST_DIR=.
-      php: 5.5
-    - env: TEST_DIR=.
-      php: 5.6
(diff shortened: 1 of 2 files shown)


carusogabriel:
@renatomefi I got the first version of this. Stages are working as expected, and specifics scripts were put in the right place. Creating a Build Stage only for `Composer lowest` isn't viable.

About the failing test for PHP 7.2: we should work on #232, to bump our version and support `PHPUnit 6`.

renatomefi:
@carusogabriel 

Thanks a lot! 
I think the advantage of having the prefer lowest separately is that we can test just a set of versions, like 5.3, 5.6 and 7.2. Now the matrix is too big and Travis will take a while!
What do you think?

I'm gonna merge #232 on branch 1.5

carusogabriel:
Interesting, didn’t think on that 🤔
Gonna create a “Composer lowest” soo 👨🏻‍💻

Issue #239, opened 2017-12-12
Does not open stream with code that generate salt by file_get_contents('/dev/urandom', false, null, 0, 16)


penja:
Hi.
Thanks for this awesome library.

When using PHP VCR with code that use `file_get_contents('/dev/urandom', false, null, 0, 16)` for generating pseudo random salt. VCR recorder do not open the stream. We have got error like this:
``` 
file_get_contents(/dev/urandom): failed to open stream: "VCR\Util\StreamProcessor::stream_open" call failed
```

killua99:
Did you fix this issue?

renatomefi:
Please update to v1.4.3

sithara:
This seems to be still breaking for me when I try to open a local file using file_get_contents in my `phpunit` tests. Appreciate any help

renatomefi:
@sithara what's the version you have installed? The version 1.4.3 is still marked as pre-release for testing, so you might have to force it on you composer.json

sithara:
My `php-vcr` version is `1.3.2` . I'm also getting very weird errors like https://github.com/php-vcr/php-vcr/issues/251 . It works correctly if its the cassette is set to record file_get_contents('http://www.google.com') . In my case, I have a `curl_exec` posting file contents to a server. Not sure where I'm going wrong

renatomefi:
@sithara so please first update to 1.4.3 and try again

PR #245, opened 2018-03-23
Stop failing on calls to /dev/urandom

alnorth:
### Context

`StreamProcessor->stream_open` calls `is_file` to check whether a stream should be intercepted. However, `is_file` returns `false` for /dev/urandom as it's not a regular file. The better check here seems to be `file_exists`.

I've added a test to check it's working. As trying to open /dev/urandom won't work on Windows I've set this test to be skipped on Windows. It's possible there's some other environment that doesn't support /dev/urandom, so worth a think.


calvinl:
+1 to this. I'm having the same issue as in #239.

Issue #201, opened 2017-06-09
Not implemented: CURLSSLOPT_NO_REVOKE 


tiagobrito:
I'm getting `Not implemented: CURLSSLOPT_NO_REVOKE (2)`
```
string(11488) "{"error":{"code":500,"message":"Internal Server Error","exception":[{"message":"Not implemented: CURLSSLOPT_NO_REVOKE (2) ","class":"BadMethodCallException","trace":[{"namespace":"","short_class":"","class":"","type":"","function":"","file":"\/php-vcr\/php-vcr\/src\/VCR\/Util\/CurlHelper.php","line":125,"args":[]},{"namespace":"VCR\\Util","short_class":"CurlHelper","class":"VCR\\Util\\CurlHelper","type":"::","function":"getCurlOptionFromResponse"
```

adri:
Thank you! What library do you use? Do you have some example code?

tiagobrito:
Hello @adri! Thanks for your reply.
I'm using PayPal  SDK for PHP.
Yes, I can provide an example later today.

tiagobrito:
@adri For example getting a payment details:
```
$result = Payment::get($id, $apiContext);
```

example: http://paypal.github.io/PayPal-PHP-SDK/sample/doc/payments/GetPayment.html

tiagobrito:
@adri Do you have any news on this ?

cerw:
Same problem with Curl/Curl

higidi:
Closing as stalled.

This was reported against the PayPal PHP SDK, which has since been replaced by the PayPal Checkout SDK (Guzzle-based). No engagement since 2017.

The underlying cause — `CURLOPT_SSL_OPTIONS` not being handled in `CurlHelper::getCurlOptionFromResponse` — is still present. If you hit this on a supported PHP version (8.0–8.5) with a current HTTP client, please open a new issue and a PR adding the missing option to the allowlist is very welcome.

PR #261, opened 2018-11-19
Add support for CURLINFO_HEADER_OUT

alnorth:
### Context

When `curl_setopt($ch, CURLINFO_HEADER_OUT, true)` has been called before a request is sent you should then be able to call `curl_getinfo($ch, CURLINFO_HEADER_OUT)` after the request to get a description of the request that's been sent. Prior to this patch calling `curl_getinfo($ch, CURLINFO_HEADER_OUT)` with VCR enabled would actually cause an exception. To compound the issue the error message was also quite confusing (see https://github.com/php-vcr/php-vcr/issues/201 for an example).

### What has been done

I've taken a second run at solving this as it was still causing problems for me with the PayPal SDK. My previous attempt was https://github.com/php-vcr/php-vcr/pull/198. I have similar misgivings about this version as it does break down some separation of concerns, but I think that's inevitable with this. The response object really does have to know about the request object.

This also required a function to format the Request as a string. I've mimicked curl's output for a simple GET request, but doing it for a POST with a body would take much more work. It's got enough there fulfil my purposes at the moment. It feels like the sort of thing that there may be a standard PHP function out there to do, but I haven't found one.

### How to test

Unit tests have been added.

### DIFF
--- a/src/VCR/Cassette.php
+++ b/src/VCR/Cassette.php
@@ -71,7 +71,7 @@ class Cassette
         foreach ($this->storage as $recording) {
             $storedRequest = Request::fromArray($recording['request']);
             if ($storedRequest->matches($request, $this->getRequestMatchers())) {
-                return Response::fromArray($recording['response']);
+                return Response::fromArray($recording['response'], $storedRequest);
             }
         }
 
--- a/src/VCR/Response.php
+++ b/src/VCR/Response.php
@@ -32,18 +32,21 @@ class Response
 
     protected $httpVersion;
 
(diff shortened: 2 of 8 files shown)


higidi:
Closing — already on master.

`CURLINFO_HEADER_OUT` is registered in `CurlHelper::$curlInfoList` (maps to `'request_header'`) and handled in `getDefaultCurlInfo()`. The feature this PR added is already available. Thanks for the contribution — the implementation landed via a different path.

PR #267, opened 2019-01-09
Add a return value

lpradovera:
Some libraries depend on this function returning a value.

### Context
I was mocking requests to [Twilio](https://github.com/twilio/twilio-php) and I realized that that library depends on `curl_setopt_array` returning something.

### What has been done
Added a simple return value.


higidi:
Thank you for reporting this and for the fix, @lpradovera.

Picked this up in #481 / #482: the return value now also mirrors the native function's
short-circuit-on-failure behavior (returns `false` as soon as one option fails, instead of always
returning `true`). Closing this in favor of #482, which builds on your fix.

Issue #268, opened 2019-02-01
Abstract Code Transform Splits In Middle Of File


bassrock:
When trying to use PHP-vcr with the stripe api it turns out that the [filter function](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/CodeTransform/AbstractCodeTransform.php#L50) happens to split the code file right in the middle of the curl statement at this line: 

https://github.com/stripe/stripe-php/blob/master/lib/HttpClient/CurlClient.php#L253

This causes php-vcr to not properly replace the curl_setops_array function and hence causing php-vcr to fail for the stripe api.  I tried a few different ways to fix this but they felt like hacks and you can not currnetly add your own code transformers.

higidi:
See https://github.com/php-vcr/php-vcr/pull/269#issuecomment-1355137112

PR #269, opened 2019-02-01
Large Files Splitting Strings

bassrock:
### Context

For main context see: #268  but the TL;DR is that the code was getting piped into the processor in pieces and was conveniently getting split right on a curl function. 
### What has been done

- This ingests the whole file before running a regex on the strings.
-

### How to test

- All tests should pass normally if you want a large file trying processing against Stripe's CurlClient.

### Todo

- [ ]
- [ ]

### Notes

-
-

### DIFF
--- a/src/VCR/CodeTransform/AbstractCodeTransform.php
+++ b/src/VCR/CodeTransform/AbstractCodeTransform.php
@@ -11,6 +11,12 @@ abstract class AbstractCodeTransform extends \PHP_User_Filter
 {
     const NAME = 'vcr_abstract_filter';
 
+    /** @var string The code to transform */
+    private  $data = "";
+
+    /** @var stdClass The clast bucket we operated on */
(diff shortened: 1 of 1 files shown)


JeroenVanOort:
I know it was quite some time ago that you submitted this PR and I'm sorry for not getting back to you earlier.

A few questions:
1. Do I understand correctly when I think that the problem you're solving here is that the code transformer can't handle files that are larger than the buffer?
2. Would you mind adding a test for this change? 
3. Would you mind rebasing your branch upon current master?

If you don't want to work on this any more, I understand. If so, please close this PR so it is clear that you'll no longer work on it.

higidi:
Ran into the same problem with a library own test case - took a while until i found out where is the problem. For now i just added some dummy comments to the file, so that buffer split do not cut the curl functions and the CurlTransformer can do his job as expected.

PR #273, opened 2019-03-25
Never validate curl post settings for non post requests

Toflar:
After my latest dependency update I suddenly get `To set a CURLOPT_READFUNCTION, CURLOPT_INFILESIZE must be set.` exceptions even though I send a `GET` request.
Something must have changed here but I guess it's just save to always early return if the request is not a `POST` request anyway because that's essentially what the method is for 😄

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -197,6 +197,10 @@ class CurlHelper
      */
     public static function validateCurlPOSTBody(Request $request, $curlHandle = null)
     {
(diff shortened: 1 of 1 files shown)


higidi:
Closing as stalled.

`CurlHelper::validateCurlPOSTBody` already early-returns when `CURLOPT_READFUNCTION` is null (line 291) and when the request body is already set (line 297). The Guzzle 6 scenario that triggered this PR is no longer an issue on the supported stack (Guzzle 7).

No activity since 2017.

Issue #282, opened 2019-10-24
SoapCodeTransform


janvernieuwe:
When there is a class that extends for example a `MySoapClientBuilder` it will be replaced with  `extends \VCR\Util\SoapClientBuilder` which doesn't exist and causes the client to be untestable while using php-vcr.

This happens in the `SoapCodeTransform`, I think the second pattern regex could be improved to prevent this kind of behavior.

JeroenVanOort:
Thank you for your contribution! I've just merged the PR.

PR #283, opened 2019-10-25
Update SoapCodeTransform regex

janvernieuwe:
### Context
Fixes issue #282 

### What has been done

- Added a failing test to see that Builders / factories and such don't get replaced
- Added a word boundary to te regex to make it work

### How to test

- Test is included

### Notes

This makes everything that extends some SoapClientBuilder or Factory testable but now only things that extend SoapClient will be replaced.


janvernieuwe:
Is there any chance to get things merged?
I made a few (huge) performance improvements on my own fork when having multiple requests in the same file with large bodies but i'm not sure if i should bother to PR them?

PR #285, opened 2019-10-28
Add CURLINFO_PRIVATE to CurlHelper

clemherreman:
## Context

This PR adds the handling of `CURLINFO_PRIVATE` in PHP VCR. This header is used by the symfony/http-client, and make it incompatible with PHP VCR.

There was a previous PR with such a fix (#278) but it was failing for PHP < 7.

### What has been done

- We added support for `CURLINFO_PRIVATE` 

### How to test

- The test harness should handle it.
- To manually test, use PHP VCR with symfony/http-client

### DIFF
--- a/src/VCR/LibraryHooks/CurlHook.php
+++ b/src/VCR/LibraryHooks/CurlHook.php
@@ -296,6 +296,14 @@ class CurlHook implements LibraryHook
      */
     public static function curlGetinfo($curlHandle, $option = 0)
     {
+        // Workaround for CURLINFO_PRIVATE.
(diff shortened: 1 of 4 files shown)


clemherreman:
Update: we actually fixed it, instead of a simple workaround.
Also we added a unit test case :)

higidi:
The fix this PR proposed — reading `CURLINFO_PRIVATE` from `$curlOptions` when no response is available yet — was independently reimplemented in #440 with a slightly cleaner approach: the merged version always reads from `$curlOptions` regardless of response state, removing the condition check. The accompanying test scenarios are also covered by #440's expanded test suite.

Closing as superseded.

PR #286, opened 2019-11-07
Add default return value for CURLINFO_CERTINFO

clemherreman:
### Context

symfony/http-client is using the cURL `CURLINFO_CERTINFO` (see [here](https://github.com/symfony/http-client/blob/master/Response/CurlResponse.php#L371))

However PHP-VCR is not handling this. By default, its value is `[]`, unless a developer use this:

```
curl_setopt($ch, CURLOPT_CERTINFO, 1);
curl_setopt($ch, CURLOPT_VERBOSE, 1);
``` 

### What has been done

This PR allows `curl_getinfo($ch, CURLINFO_CERTINFO)` to return the default cURL value : `[]` when those `curl_setopt` are not set, instead of a PHP VCR Exception.

### How to test

I added a test to `CurlHelperTest`, testing that calling `curl_getinfo($ch, CURLINFO_CERTINFO)` doesn't thrown an exception, but returns the default cURL value.

### Notes

Build is failing because of PHP CS Fixer on files unrelated to this PR. As multiple PRs include that fix, I didn't run the fix again on this one :)

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -111,6 +111,10 @@ class CurlHelper
             case CURLINFO_HEADER_SIZE:
                 $info =  mb_strlen(HttpUtil::formatAsStatusWithHeadersString($response), 'ISO-8859-1');
                 break;
+            case CURLINFO_CERTINFO:
+                $curlInfo = $response->getCurlInfo($option);
+                $info = (!is_null($curlInfo)) ? $curlInfo : [];
+                break;
             default:
                 $info = $response->getCurlInfo($option);
                 break;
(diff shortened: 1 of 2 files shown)


higidi:
This has been adopted in PR #435 (now merged), which preserves your original
commit (476d172) with author attribution.

A follow-up commit in the same PR corrected two latent issues in the original
patch:
- `Response::getCurlInfo()` is keyed by string name; the original passed the
  integer constant (`$option`) which would always have resolved to `null`.
- `CURLINFO_CERTINFO` was missing from `CurlHelper::$curlInfoList`, so
  `curl_getinfo($ch)` (all options) wouldn't have included it.

Thanks for the patch, @clemherreman — sorry it took so long!

PR #288, opened 2019-11-28
Increase performance by storing response body separately

janvernieuwe:
### Context

If you have a file with multiple large multi line response bodies, it will cause php-vcr to become very slow

### What has been done

- On storing the requests, looks for the response body and puts it in a separate file
- On loading, it will test if the body has been replaced with a file reference, and replaces the placeholder with the content fetched from the file

### How to test

- Make a test that requests multiple large response bodies

### Todo

- improve detection of placeholder?
- This is basically a RFC PR, not per se final

### Notes

- caused a functional test of mine to go from 30s to 0.5s

### DIFF
--- a/src/VCR/Storage/AbstractStorage.php
+++ b/src/VCR/Storage/AbstractStorage.php
@@ -118,4 +118,26 @@ abstract class AbstractStorage implements Storage
             fclose($this->handle);
         }
     }
+
+    protected function storeBody(&$recording)
+    {
+        if (!isset($recording['response']['body'])) {
+            return;
+        }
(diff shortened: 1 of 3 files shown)


joelcollinsdc:
just wanted to point out.... on the master branch (merged in 2018, no releases yet?) there is a perf improvement on the YAML storage that significantly helps out.  this should be ported over to JSON?

janvernieuwe:
I have not tested it, is the performance increase so huge?

My approach is different tho, by storing the response body in a separate file, you don't need to loop over the many lines of a large multi line response.

Perhaps a combination of both would be even faster :)

PR #294, opened 2020-01-24
[Curl] Add \r\n to HTTP status and headers to mimic what curl does

clemherreman:
### Context

When returning the array of cURL raw headers, PHP-VCR returns them like this

```php
[
    "Content-Type: text/html; charset=UTF-8",
    "Content-Encoding: gzip",
]
```

However, cURL returns them more "raw", with the trailing `\r\n` mandatory at the end of every HTTP header line:

```php
[
    "Content-Type: text/html; charset=UTF-8\r\n",
    "Content-Encoding: gzip\r\n",
]
```

This PR changes the PHP-VCR behavior. This is necessary to HTTP clients, like Symfony HTTP Client that rely on those `\r\n` to [detect the end of the headers and the beginning of the request body, when parsing it.](https://github.com/symfony/http-client/blob/b42500787e4bbca335af4ed0fe377462fefc3733/Response/CurlResponse.php#L300) ([or just to extract it via `substr()`](https://github.com/symfony/http-client/blob/b42500787e4bbca335af4ed0fe377462fefc3733/Response/CurlResponse.php#L302))

### What has been done

`VCR/Util/HttpUtil:formatHeadersForCurl()` and `VCR/Util/HttpUtil:formatAsStatusString()` now returns a trailing `\r\n` to the HTTP status and HTTP header strings. 

This is the actual PHP cURL behavior. 

Note: **This could, however be a BC change** if someone was relying on not having those `\r\n`. However, it would then break in actual cURL requests, so I doubt that it is the case. 

### How to test

Unit tests were updated.

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -59,7 +59,7 @@ class CurlHelper
         if (isset($curlOptions[CURLOPT_HEADERFUNCTION])) {
             $headerList = array(HttpUtil::formatAsStatusString($response));
             $headerList = array_merge($headerList, HttpUtil::formatHeadersForCurl($response->getHeaders()));
-            $headerList[] = '';
+            $headerList[] = "\r\n";
             foreach ($headerList as $header) {
                 call_user_func($curlOptions[CURLOPT_HEADERFUNCTION], $ch, $header);
             }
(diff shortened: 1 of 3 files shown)


higidi:
This has been adopted and merged via #433. Thank you @clemherreman for the original fix — the commit is preserved with your authorship.

Issue #289, opened 2019-12-02
PHP7.4 & guzzle


janvernieuwe:
Since upgrading to php7.4 im getting

```
ParseError : syntax error, unexpected end of file, expecting :: (T_PAAMAYIM_NEKUDOTAYIM)
 vendor/guzzlehttp/guzzle/src/Handler/CurlMultiHandler.php:197
```
I have noticed this on multiple projects. I'll try to look into this later today, but as of now it's not really usable on php 7.4.

I only get this error when VCR is turned on.

JapSeyz:
Did you dig further into this?

janvernieuwe:
I tried, but there is some really weird issue going on. It almost seems that files are not completely loaded after x lines. It's very strange. If I find something, I'll make it known here.

brunowowk:
Maybe it's a regression on this php 7.4 bug: https://bugs.php.net/bug.php?id=78406

Tried finding the exact spot where it occurs with xdebug and it seems to happen when the file is included.
The CodeTransform works as expected, but somehow it seems that part of the stream is ignored, probably because the code transform increases the file size and the old file size is considered.

Looks like It's being reproduced on travis on the 7.4 test suite on #290:
https://travis-ci.org/php-vcr/php-vcr/jobs/619582578?utm_medium=notification&utm_source=github_status

Just by having this on the bootstrap it breaks the test suite:
```
\VCR\VCR::turnOn();
\VCR\VCR::turnOff();
```
When commenting it the tests run, breaking in other spots, and reporting several errors of this kind when composer tries to autoload some classes:
```
ParseError: syntax error, unexpected end of file in php-vcr/src/VCR/Util/HttpClient.php:35
Stack trace:

#0 php-vcr/vendor/composer/ClassLoader.php(322): Composer\Autoload\includeFile(...
```

By commenting `$this->enableLibraryHooks();` on `\VCR\Videorecorder::turnOn()` I get no more syntax errors. So it's most definitely a problem on the stream processing used by those hooks

It's either a bug on core or some undocumented change on streams that broke the stream processing.
The only related change I could find was: https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.stream-wrappers 
But that doesn't seem to explain what we see going on here.

morozov:
[Bug #78406](https://bugs.php.net/bug.php?id=78406) is closed and according to https://3v4l.org/jEMEE is no longer reproducible although underlying logic and the symptoms look very similar.

[...]

PR #293, opened 2020-01-06
Fix the parse error on PHP 7.4

morozov:
Do not return the stat since we don't know the resulting the size that the file will have after having all transformations applied. When including files, PHP 7.4 and newer are sensitive to file size reported by stat.

Fixes #289 (based on the recommendation from https://bugs.php.net/bug.php?id=79072).

PR #295, opened 2020-01-31
BUGFIX: ErrorException in VCR/Storage/Yaml.php:61

bomas13:
### Context
if the fixture is recently created and it has no content, the parser returns `null` to `$recording`. And array-like access to `null` via `$recording[0]` will raise an exception. With this little check this behavior is fixed and also all tests will work properly.

### What has been done

- added a check if $response is null

### How to test

- write a simple vcr test and try to record a whole new fixture

### Todo

- merge into php-vcr/php-vcr:master and tag a new version

### Notes

- thanks to @morozov who fixed the streamfilter bug
- thanks to @danielbecker who helped debugging this issue
- see https://github.com/php-vcr/php-vcr/issues/289

### DIFF
--- a/src/VCR/Storage/Yaml.php
+++ b/src/VCR/Storage/Yaml.php
@@ -58,8 +58,10 @@ class Yaml extends AbstractStorage
     public function next()
     {
         $recording = $this->yamlParser->parse($this->readNextRecord());
-        $this->current = $recording[0];
-        ++$this->position;
+        if (!is_null($recording)) {
+            $this->current = $recording[0];
+            ++$this->position;
+        }
     }
 
     /**
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -262,13 +262,17 @@ class StreamProcessor
     /**
      * Retrieve information about a file resource.
      *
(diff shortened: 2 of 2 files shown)


bwood:
Since commit `42ec927` already has a separate PR (https://github.com/php-vcr/php-vcr/pull/293) which pre-dates this one, maybe it would be better to treat these two issues separately?

src/VCR/Storage/Yaml.php
```diff
@@ -58,8 +58,10 @@ public function storeRecording(array $recording)
     public function next()
     {
         $recording = $this->yamlParser->parse($this->readNextRecord());
-        $this->current = $recording[0];
-        ++$this->position;
+        if (!is_null($recording)) {
```
bwood, line 61: It shouldn't happen, but if `$recording = 'some string'`, unintended things will happen.   I'd suggest:
```
if (is_array($recording) && count($recording))
```

higidi:
Closing — fix is already on master.

`Storage/Yaml.php` now uses `$recording[0] ?? null` (null-coalesce operator) when reading the next record, so an empty or newly-created cassette no longer throws `ErrorException`. The fix was applied differently from this PR but achieves the same result.

PR #298, opened 2020-04-04
Php 7.4 fix

snags88:
### Context
PHP-VCR is broken with PHP v7.4

Merging in this fix https://github.com/php-vcr/php-vcr/pull/293


snags88:
Oops wrong base

Issue #272, opened 2019-03-21
HttpUtil::parseResponse fails on empty $response


kaktus-74:
It can happen that $response will be empty string (sometimes on 404 error)

and line 
`list($rawHeader, $rawBody) = explode("\r\n\r\n", $response, 2);`
will throw an exception
`ErrorException: Undefined offset: 1 in \vendor\php-vcr\php-vcr\src\VCR\Util\HttpUtil.php on line 70`

higidi:
Keeping open — this is a valid bug. An empty or body-less response (e.g. HTTP 204 No Content) causes `explode("\\r\\n\\r\\n", $response, 2)` to return only one element, resulting in an undefined offset warning on PHP 8.

PR #302 has a fix for this — contributions and review welcome.

PR #242, opened 2018-02-09
array_pad on list+explode

bucha:
This fixes some edge-cases that we encountered during unit-tests, where the response might not carry a body for example.

### What has been done

-
-

### How to test

-
-

### Todo

- [ ]
- [ ]

### Notes

-
-

PR #302, opened 2020-05-01
Prevent HttpUtil::parseResponse() from emitting PHP Notice Undefined offset 1

bwood:
### Context

```      
```

### What has been done

### How to test

### Notes


higidi:
Superseded by #476, which applies the same array_pad() fix against the current version of parseResponse() and adds regression tests compatible with the string typehint. Closing in favor of that PR.

PR #306, opened 2020-05-21
Fix typo in README and remove warning

MrRio:
Rationale: It's pretty battle-tested now, as we tidy this up we want to encourage new users.


JeroenVanOort:
Thanks!

PR #311, opened 2020-06-13
Symfony 5 compatibility

pauladams8:


JeroenVanOort:
Thank you for your work! Would you mind doing your PR to the 1.5 branch? Although github doesn't show it, the CI for this PR has failed for PHP <= 7.1: https://travis-ci.org/github/php-vcr/php-vcr/builds/697325760. These versions are no longer tested for the 1.5 branch, so it won't be a problem there. I plan on do a pre-release for 1.5.0 soon, so it can be tested in the wild. Adding compatibility with Symphony 5 is a great contribution to that.

composer.json
```diff
@@ -21,7 +21,7 @@
         "ext-curl": "*",
         "beberlei/assert": "^2.0",
         "symfony/yaml": "~2.1|^3.0|^4.0",
-        "symfony/event-dispatcher": "^2.4|^3.0|^4.0"
+        "symfony/event-dispatcher": "^5.0"
```
allejo, line 24: Is php-vcr 1.5 going to drop [Symfony 3.4 and 4.4, which are LTS releases](https://symfony.com/releases)?

src/VCR/Videorecorder.php
```diff
@@ -95,13 +95,13 @@ public function getEventDispatcher()
     /**
      * Dispatches an event to all registered listeners.
      *
-     * @param string $eventName The name of the event to dispatch.
      * @param Event $event The event to pass to the event handlers/listeners.
+     * @param string $eventName The name of the event to dispatch.
      * @return Event
      */
-    private function dispatch($eventName, Event $event = null)
+    private function dispatch(Event $event, $eventName = null)
```
pauladams8, line 102: Should I add a return type hint to match the style in the 1.5 branch?

src/VCR/Videorecorder.php
```diff
@@ -95,13 +95,13 @@ public function getEventDispatcher()
...
     {
-        return $this->getEventDispatcher()->dispatch($eventName, $event);
+        return $this->getEventDispatcher()->dispatch($event, $eventName);
```
pauladams8, line 104: If VCR is to support LTS Symfony releases, we'll need to check the version at runtime as 4.3 contains a breaking change which affects the order of parameters to the `dispatch()` method of the event dispatcher. The 1.5 branch currently uses some dirty reflection to determine the order. See https://github.com/php-vcr/php-vcr/commit/71372309fc35d22223085fe65d95516cc2e6b19d.

composer.json
JeroenVanOort: I think we should indeed continue to support Symfony 3.4 and 4.4. Thank you for asking that question!

src/VCR/Videorecorder.php
JeroenVanOort: If possible, yes please.

src/VCR/Videorecorder.php
JeroenVanOort: I can't find a version constant or something else we can check during runtime to determine the order of arguments, as alternative to the reflection solution. Do you know of an other option?

pauladams8: It's probably the most straightforward solution. However, since `Symfony\Components\EventDispatcher\Event` was removed in the same release the signature was changed, we could check for the existence of that class.

Example:
```php
private function dispatch(Event $event, $eventName = null)
{
    if (class_exists('Symfony\Components\EventDispatcher\Event') {
        return $this->getEventDispatcher()->dispatch($event, $eventName);
    }
    
    return $this->getEventDispatcher()->dispatch($eventName, $event);
}
```

Changelog: https://github.com/symfony/symfony/blob/678e066aa8b2d91bec50c56a081435300f58fcf8/src/Symfony/Component/EventDispatcher/CHANGELOG.md#500

pauladams8:
Should be backwards compatible with older Symfony LTS releases now

JeroenVanOort:
Thanks again for your work. If you could bring your branch up to speed with the current 1.5 (which had your other PR merged with it), I'm ready to merge.

Issue #312, opened 2020-06-15
[Bug] Malformed Set-Cookie header


pauladams8:
VCR's HTTP client is incorrectly folding `Set-Cookie` headers on the response into a single, comma separated header. This practice was deprecated as of April 2011 under RFC 6265 due to parsing conflicts with the `,` character.

To quote the RFC:

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.

I've spent a while pinning this down, as it was preventing an HTTP client my application uses (Guzzle) from parsing the cookies properly, which in turn led to authentication issues in my tests.

pauladams8:
See https://github.com/php-vcr/php-vcr/blob/267674dc88e0dbe304c6de79697215e295a36d05/src/VCR/Util/HttpUtil.php#L15

PR #313, opened 2020-06-15
Changed header parsing logic

pauladams8:
See issue #312

### DIFF
--- a/src/VCR/Request.php
+++ b/src/VCR/Request.php
@@ -171,6 +171,10 @@ class Request
      */
     public function getHeader($key)
     {
+        if (!isset($this->headers[$key])) {
+            return null;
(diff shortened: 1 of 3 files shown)


pauladams8:
Should probably have explained this better. I've changed the way VCR parses headers so instead of folding multiple header values into a single comma separated header, it returns an array of the values.

Not sure if this is the best way of parsing headers, might be worth doing some research but I think it is the most intuitive.

JeroenVanOort:
Would you mind fixing the cs-fixer problem as reported here? https://travis-ci.org/github/php-vcr/php-vcr/jobs/698893846

When you do the commit for that, github should report the CI check for this PR correctly.

PR #321, opened 2020-07-21
StreamProcessor stream_stat to only return false for filtered streams

endrsmar:
### Context
Using php-vcr alongside fpdi causes issues for us because of the way `StreamProcessor` alters result of `\stat` calls.

### What has been done
Currently `StreamProcessor::stream_stat` always returns false even for streams that are not actually being altered (no code transformations attached to the stream). This PR adds a check to only fail the `stream_stat` call if the stream is actually altered.

### How to test
Test has been added to the `StreamProcessorTest` suite.

### DIFF
--- a/src/VCR/Util/StreamProcessor.php
+++ b/src/VCR/Util/StreamProcessor.php
@@ -54,6 +54,11 @@ class StreamProcessor
     protected $isIntercepting = false;
 
     /**
+     * @var bool True if any code transformer has been attached to the stream, false otherwise
+     */
+    protected $isTransforming = false;
+
+    /**
      *
      * @param Configuration $configuration
      */
@@ -272,7 +277,11 @@ class StreamProcessor
      */
     public function stream_stat()
     {
-        return false;
+        if ($this->isTransforming) {
+            return false;
+        }
+
+        return \fstat($this->resource);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


JeroenVanOort:
Thank you for your contribution!

I think I've already fixed this on the 1.5 branch: https://github.com/php-vcr/php-vcr/pull/317/files . Could you please check if that solution works for you?

src/VCR/Util/StreamProcessor.php
```diff
@@ -629,9 +638,13 @@ public function detachCodeTransformer(AbstractCodeTransform $codeTransformer)
      */
     protected function appendFiltersToStream($stream)
     {
+        if (\count(static::$codeTransformers) === 0) {
+            return;
+        }
         foreach (static::$codeTransformers as $codeTransformer) {
             stream_filter_append($stream, $codeTransformer::NAME, STREAM_FILTER_READ);
         }
+        $this->isTransforming = true;
```
JeroenVanOort, line 647: I think I would implement this like this:

```php
if (!empty(static::$codeTransformers)) {
    $this->isTransforming = true;
}
```

In my opinion that would be more readable.

endrsmar:
Hey, yeah the fix in 1.5 works great, thanks. :+1: Didn't know that master was not the "latest" in development :smile: I guess I can just close this PR then.

src/VCR/Util/StreamProcessor.php
endrsmar: Yes that would be better readable :slightly_smiling_face:

JeroenVanOort:
I'll be merging 1.5 to master very soon ;-)

PR #334, opened 2020-11-23
Support for ssl requests in some edge cases like paypal php sdk

specialtactics:
### Context
When using SSL requests, for example with paypal php sdk, it fails to go through vcr properly

### What has been done
Added this one option to getCurlOptionFromResponse

### How to test
Not exactly sure, but for example the issue for us happens in PayPal\Api\Sale::refundSale()

### Notes
Added a define for pre - php 7.3, for the curl constant

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -2,6 +2,8 @@ parameters:
     level: 7
     paths:
         - src
+    # Can remove reportUnmatchedIgnoredErrors option after php 7.2 is no longer supported
+    reportUnmatchedIgnoredErrors: false
     ignoreErrors:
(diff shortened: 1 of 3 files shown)


specialtactics:
Just a note, the curl constant used only exists as of php 7.3, so I had to do a conditional define for it for php 7.2, and phpstan doesn't pick this up. I read about it on the phpstan wiki, and for this limitation, the best thing is to ignore it, so I added it to the ignore list.

In both cases, I've commented to remove those things once we are no longer supporting php 7.2.

specialtactics:
Thanks for your promptness! Let me know when it's tagged.

PR #344, opened 2021-03-21
add redactions to VCR

aksonnic:
### Context

Typically there are lots of private values in network request/response that should not be saved to storage, and potentially be committed as part of a cassette. This PR allows users to specify redactions to be applied before/after storage. In all other respects the library works as it did before.

### What has been done

- added ability to specify redactions to be applied
- added scrubbing/unscrubbing just before a recording is recorded/played back
- tests

### How to test

- run unit tests

### Notes

- I don't really like the asymmetry between `scrub($request, $response)` and `unscrub($recording)`. It's like that because each direction instantiates `Request` or `Response` objects at different times, and was an attempt to prevent needless array<->object conversions, but maybe it's not worth bothering.
- `Scrubber`/`scrub`/`unscrub` perhaps isn't the best naming

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Disclaimer: Doing this in PHP is not as easy as in programming languages which s
   implement a custom request matcher to handle any need.
 * The recorded requests and responses are stored on disk in a serialization format of your choice
   (currently YAML and JSON are built in, and you can easily implement your own custom serializer)
+* Private data can be redacted from storage using `addRedaction()`
 * Supports PHPUnit annotations.
 
 ## Usage example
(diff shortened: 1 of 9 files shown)


Justintime50:
Can we get eyes on this? This is an absolute must for security and a blocker before we could use it on our projects.

TheFehr:
@aksonnic I added your changes to my own fork, as I also needed a fix for a small curl bug. Can you give me an example of how to use it? I see you added tests, but there you create your own Cassettes. Is that the only way? Is this no compatibly with the "default" use of 

```php
VCR\VCR::turnOn();
VCR\VCR::insertCassette('some_name.yml');

/** The tested code **/

VCR::eject();
VCR::turnOff();
```

Thanks a lot

Issue #349, opened 2021-05-12
Error: Could not init curl with URL ... (Guzzle + VCR)


Justintime50:
# Summary

I'm attempting to record cassettes with VCR and Guzzle and am getting the following error: `Could not init curl with URL ...` which is [thrown here](https://github.com/php-vcr/php-vcr/blob/989fdcad482d830890757b8165127ed0183de41b/src/VCR/Util/HttpClient.php#L26). After hours of trial and error, I decided to simply remove this assertion from the vendor VCR library and it happily recorded my cassette.

I was most likely doing something wrong, but in the off chance I wasn't and there is indeed a bug here, I wanted to report it so we can take a look into fixing it.

## Additional Info

Here are some code snippets from my project:

```php
# CustomersTest.php

<?php

use PHPUnit\Framework\TestCase;

class CustomersTest extends TestCase
{
    /**
     * Test creating a customer
     *
     * @vcr create-customer.yml
     * @return void
     */
    public function testCreate()
    {
        $client = new TuneupTechnology\Client(getenv("API_EMAIL"), getenv("API_KEY"), 'http://tuneapp.localhost/api');

        $customer = $client->customers->create(
            $data = [
                "firstname"     => "Jake",
                "lastname"      => "Peralta",
                "email"         => "jake@example.com",
                "phone"         => "8015551234",
                "user_id"       => 1,
                "notes"         => "Believes he is a good detective.",
                "location_id"   => 2,
            ]
        );

        $this->assertEquals($customer->getStatusCode(), 200);
    }
}
```

```php
# bootstrap.php

<?php

use VCR\VCR;

VCR::configure()->setCassettePath('test/cassettes')
    ->setWhiteList(array('vendor/guzzle'))
    ->setStorage('yaml')
    ->setMode('once');
```

```php
# Client.php

    public function makeHttpRequest($method, $endpoint, $data = null)
    {
        try {
            $guzzle_client = new \GuzzleHttp\Client();

            $headers = [
                "Accept" => "application/json",
                "User-Agent" => "TuneupTechnologyApp/PHPClient/$this->version",
                "Email" => $this->email,
                "Api-Key" => $this->api_key,
            ];

            $response = $guzzle_client->request(strtoupper($method), $endpoint, [
                'headers' => $headers,
                'timeout' => $this->timeout,
                'json' => $data,
            ]);
        } catch (\GuzzleHttp\Exception\ClientException $error) {
            return $error->getResponse()->getBody();
[...]

PR #351, opened 2021-05-16
Make curl_init assertions PHP 8 compatible

dbrumann:
### Context

With PHP 8 `curl_init()` does no longer return a `resource` and instead returns an object (`CurlHandle`/`CurlMultiHandle`).

### What has been done

- Adds PHP 8 support for HttpClient
- Fixes test suite for PHP 8
- Adds PHP 8 to Travis CI matrix


Justintime50:
I'm interested to see this gain PHP8 support as I found it wasn't compatible via #349. Will have to downgrade for now.

dbrumann:
@Justintime50 changing the assertion in the HttpClient fixes it for me on PHP 8.0.5.

If all else fails, you could fork my branch and then use that as a custom repository:

```
# composer.json
{
    "repositories": [
        {
            "type": "vcs".
            "url": "https://github.com/dbrumann/php-vcr" # repalce with your fork in case my branch goes away after merge
        }
    ].
    "require-dev": {
        "php-vcr/php-vcr": "dev-curl_replace_assertion"
        # ...
    }
    # ...
}
```

I am not sure how to handle the failing Travis build. It will be hard to satisfy phpstan without ignoring errors and `reportUnmatchedIgnoredErrors: false` because errors differ in PHP 7/8. Can anyone from the core team advise on how to proceed with this?

dbrumann:
Can anyone from @php-vcr advise on how to proceed with this, especially the phpstan rules? I would really like to make the curlHook compatible with PHP 8

JeroenVanOort:
The only thing I can think of right now, is asserting a variable != false, instead of asserting it is a resource.

JeroenVanOort:
I've been looking into adding compatibility for PHP 8 for about an hour now and I'm starting to think that it might be good to drop support for PHP7 entirely. In order to keep PHP7, we would need to add all kinds of phpstan ignore rules for different versions. Anyone is welcome to show me that it can be done without making a mess, but you're just as welcome to put some work into dropping PHP8.

dbrumann:
> we would need to add all kinds of phpstan ignore rules for different versions

Yes, I agree. This is exactly where I got stuck. If dropping PHP 7 support is an option, I will gladly update the PR accordingly in the next few days.

ecourtial:
Great news ! I hope that we won't have any BC break :stuck_out_tongue:

JeroenVanOort:
I have just released version 1.6.0, which is to be used with PHP8.

PR #354, opened 2021-08-02
Fix PHP8 curl_init() return value

igaster:
### What has been done

-
-

### How to test

-
-

### Todo

- [ ]
- [ ]

### Notes

-
-


Dwarfex:
As i run into the same issue as here: Isn't this a less complete solution as https://github.com/php-vcr/php-vcr/pull/351 ?

JeroenVanOort:
I have just released version 1.6.0, which is to be used with PHP8.

Issue #336, opened 2021-01-19
Incompatibility with symfony/event-dispatcher


zebraf1:
Using php-vcr 1.5 conflicts with symfony/event-dispatcher 5.0

requires symfony/event-dispatcher: ^2.4|^3.0|^4.0|^5.0

** How to reproduce: **
* install php-vcr ^1.5
* install symfony/event-dispatcher ^5.0

 
[TypeError] Argument 1 passed to Symfony\Component\EventDispatcher\EventDispatcher::dispatch() must be an object, string given, called in .../vendor/php-vcr/php-vcr/src/VCR/Videorecorder.php on line 99

* https://github.com/symfony/event-dispatcher/blob/5.0/CHANGELOG.md#430 (dispatch method signature changed and raises deprecation warning)
* https://github.com/symfony/event-dispatcher/blob/5.0/CHANGELOG.md#500
The signature of the EventDispatcherInterface::dispatch() method has been changed to dispatch($event, string $eventName = null): object.

**Problem:**
dispatch is called with (string, Event) types.
https://github.com/symfony/event-dispatcher/blob/5.0/EventDispatcher.php#L48 - this new interface requires first parameter to be object.

**Solution**:
* Use appropriate way to check whether EventDispatcher is using old or new interface
* Do not rely on class_exists(\Symfony\Component\EventDispatcher\Event::class)

zebraf1:
https://github.com/php-vcr/php-vcr/commit/3d88f89ae2f915471043ebe61ff964937bc246d6#diff-8f9fb128fc0a304a8c6dcd7ba61771a13170c368a48a9fabf993fc097416469fR104

This code should make the code compatible with 5.0 also. So it is weird that my project finds \Symfony\Component\EventDispatcher\Event::class exists

zebraf1:
If you use codeception/codeception ^4.1.15 then you will have Event::class at that namespace due to this change:

https://github.com/Codeception/Codeception/commit/15ba342b6511d0557435182d815690aa6b8e6aef
```php
//Compatibility with Symfony 5
if (!class_exists('Symfony\Component\EventDispatcher\Event') && class_exists('Symfony\Contracts\EventDispatcher\Event')) {
    class_alias('Symfony\Contracts\EventDispatcher\Event', 'Symfony\Component\EventDispatcher\Event');
}
```

zebraf1:
[...]

PR #358, opened 2021-10-08
Proper Detection of EventDispatcherInterface

zebraf1:
Fix issue #336

### Context
dispatch() was called with wrong order of parameters when using Codeception 4.1.15 or higher which added a class alias for EventDispatcher\Event.

### What has been done
- Using recommendation by @naktibalda to detect which version of dispatch method must be used.

### How to test

- install codeception/codeception ^4.1.14
- install php-vcr ^1.5
- install symfony/event-dispatcher ^5.0
- Call Videorecorder->handleRequest() with proper cassette inserted
- Must not produce error: [TypeError] Argument 1 passed to Symfony\Component\EventDispatcher\EventDispatcher::dispatch() must be an object, string given, called in .../vendor/php-vcr/php-vcr/src/VCR/Videorecorder.php on line 99


Naktibalda:
There is no point to merge this change to master, because master no longer supports symfony/event-dispatcher v4.
You could target 1.5 branch instead.

zebraf1:
Oh yeah, but then I need to remove all the commits not in 1.5.
I moved this commit to 1.5 branch and created a separate PR #365 because I cannot change the source branch of a PR, only the target branch.

PR #362, opened 2022-02-21
Update HttpUtil responseParsing to handle multiple HTTP responses

sf-cg:
### Context

Within our projects we use a proxy when connecting to some services, and when doing this we get an extra HTTP/... in the response string that is then parsed. Similar to the 100 continue issue, but ours were 200 OK response from the proxy added ontop

### What has been done

- the HttpUtil parsing of response has been update to only get the headers after the last HTTP/ line, which ideally should make this much more flexible than just stripping any 100 continues
- test added to make sure cases other than 100 continue are treated correctly

### How to test

- If possible use a proxy that adds a response (ideally not 100 continue)
- check the automated test that was added

### DIFF
--- a/src/VCR/Util/HttpUtil.php
+++ b/src/VCR/Util/HttpUtil.php
@@ -68,9 +68,20 @@ class HttpUtil
      */
     public static function parseResponse(string $response): array
     {
-        $response = str_replace("HTTP/1.1 100 Continue\r\n\r\n", '', $response);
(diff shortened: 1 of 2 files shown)


specialtactics:
Good work!

higidi:
Closing as stalled — superseded by PR #413.

PR #413 includes a more recent fix for `parseResponse` handling multiple HTTP response prefixes (proxy responses, gateway acks). It is already assigned for review. Please follow up there.

Issue #363, opened 2022-02-23
BadMethodCallException: Not implemented: CURLINFO_EFFECTIVE_URL (1048577) 


whikloj:
In a curl multihandle I use this curlinfo option to get the current handle's URL. It appears relatively easy to fix, but I'm also very new to php-vcr.

higidi:
As announced in #413 i'll create a new patch to support `CURL_EFFECTIVE_URL`.

higidi:
Ok, i didn't unterstand why we should do that. With pr #300 we also persist the curl info array (also the `url` which is to be the `effective_url`) so it should be also respond to `CURLINFO_EFFECTIVE_URL`.
What you've done is to always respond always with request url, which might differ from the effective url. So i didn't understand the benefit. Is this meant as a kind of fallback?

higidi:
Closed because no answer

PR #364, opened 2022-02-23
Pass url into curlinfo hook for effective url

whikloj:
### Context
Issue #363

When checking for errors in a `curl_multi` I pull the effective url to log with the error.
```php
$url = curl_getinfo($curl_handles[$x], CURLINFO_EFFECTIVE_URL);
```
Which currently fails with php-vcr
```
BadMethodCallException: Not implemented: CURLINFO_EFFECTIVE_URL (1048577)
```
IMHO it really should just return the request url parameter from the source YAML.

### What has been done

- Added a nullable string parameter to the `getCurlOptionFromResponse` to pass in the URL (or null)
- Get the url from the `self::$requests` array of VCR\Request objects

### How to test

- Before if you did a command like above you get the above error. 
- Now you should get the url provided in the `request -> url` parameter

### Todo

- [ ]
- [ ]

### Notes

- I'm very new to php-vcr, so this might go against the pattern you are trying to work to. If so any guidance would be appreciated. 
-

### DIFF
--- a/src/VCR/LibraryHooks/CurlHook.php
+++ b/src/VCR/LibraryHooks/CurlHook.php
@@ -281,6 +281,7 @@ class CurlHook implements LibraryHook
         if (isset(self::$responses[(int) $curlHandle])) {
             return CurlHelper::getCurlOptionFromResponse(
                 self::$responses[(int) $curlHandle],
+                self::$requests[(int) $curlHandle]->getUrl(),
                 $option
             );
         } elseif (isset(self::$lastErrors[(int) $curlHandle])) {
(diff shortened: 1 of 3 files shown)


whikloj:
I'm using the 1.5 version to support some older versions of PHP. If this PR seems acceptable, would you be open to another patch release of that version?

specialtactics:
Hi, Would invite you to make this PR on our fork, if not we will likely do so ourselves at some point:
https://github.com/CoverGenius/php-vcr

higidi:
See discussion started on #363

PR #370, opened 2022-10-28
Chore/add return types for 8 1 iterator compatibility

WillFoSho:
### Context
Need to bring package up to PHP8.1 compatibility

### What has been done
Modified various iterator implementations to match 8.1 iterator interface https://www.php.net/manual/en/class.iterator.php

### How to test
Appears to work for 8.0 and 8.1. If you wish to test in 8.1 update the dockerfile to:
`FROM php:8.1-cli-alpine`
And run tests as usual.

Note: No extra tests have been added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ tests/fixtures/cassette*
 composer.lock
 .phpunit.result.cache
 .php-cs-fixer.cache
+.idea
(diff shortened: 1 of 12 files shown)


WillFoSho:
Uh... this is not where I wanted to create this PR. Please ignore.

PR #379, opened 2022-12-19
Record identical requests

higidi:
### Context

This PR is the rebased result of #270 and small fixes and is hopefully the last try to bring this feature back to the `master`. There is nothing more to say, for more information see #161 (which inspired #270) or the issue #132.

### What has been done

- Refactored/rebased the work of #161 and #270 onto the current master
- Fixed a backward compatibility issue where the cassette did not handle legacy stored records correctly, see 96721547bb50ddee25ffbfb749b4df6dae7bc6e0

### TODO

- Port this feature back to the 1.5 branch (will be done after merging this)

### DIFF
--- a/src/VCR/Videorecorder.php
+++ b/src/VCR/Videorecorder.php
@@ -103,7 +103,9 @@ class Videorecorder
     public function eject(): void
     {
         Assertion::true($this->isOn, 'Please turn on VCR before ejecting a cassette, use: VCR::turnOn().');
+
         $this->cassette = null;
+        $this->resetIndex();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/VCR/Videorecorder.php
```diff
@@ -214,4 +226,19 @@ public function __destruct()
             $this->turnOff();
         }
     }
+
+    protected function iterateIndex(Request $request): int
+    {
+        $hash = $request->getHash();
```
dam1r89, line 232: The issue with this method is that it assumes a request is unique based on its hash, but this may not always be the case.

For instance, if someone turns off the header request matcher, requests with the same body, URL, and other parameters but different headers should still be treated as the same request and receive the sequential index.

However, this solution would result in different hashes for those identical requests due to the differing headers, and both would get the same index.

I hope this is clear enough.

Would you accept PR or does this help you to resolve it?

PR #385, opened 2023-02-03
FEATURE: Add testcase

dam1r89:
### Context

With this test case, I'm reproducing the problem that I'm facing. If the request matcher is ignoring some parts of the request making them the same, those requests should increment the same key in the `$indexTable`.

### What has been done

- Failing test explaining the problem that I have

### How to test

- Run the test

### Notes

If you are open for accepting it, I could submit a fix for this issue.

### DIFF
--- a/src/VCR/Videorecorder.php
+++ b/src/VCR/Videorecorder.php
@@ -38,9 +38,9 @@ class Videorecorder
     protected EventDispatcherInterface $eventDispatcher;
 
     /**
-     * @var array<string, int>
+     * @var Request[]
      */
-    protected array $indexTable = [];
+    protected array $handledRequests = [];
 
     public function __construct(
         protected Configuration $config,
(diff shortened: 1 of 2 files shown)


higidi:
Hi @dam1r89 

thx 4 contributing! I think i understood the problem.

What do you think about changing or implementing a new request matching api to return something like a hash (same as `Request::getHash()` does before) to still allow using of an index-table instead persisting and matching all requests again and again when a new arrives?

I'm asking because on long running processes this might get a problem.

What do you think?

dam1r89:
At the moment, I am occupied with other tasks and cannot say when I will have the time to review this. 

I  am curious about the scale of the tests you referenced - are they limited to hundreds of requests? 

It may not be necessary to pursue optimization unless it becomes a significant issue, as the majority of tests will likely not generate a large volume of requests.

dam1r89:
Hey @higidi, any update on this? Just to let you know, we are currently using the version with this patch without any problems on a fairly large code base.

I wanted to check in to see if you would be open to accepting another PR with an additional feature that I need.

Issue #373, opened 2022-12-15
Running the test suites will be failed with PHP 8.2 version


peter279k:
As title, it seems that the test suites will be failed with PHP 8.2 version. Here are steps to reproduce:

- Installing the `PHP 8.2` version.
- Cloning the the `php-vcr` repository. And I ensure latest stable version also has this issue.
- Installing the dependencies via the `composer install` command.
- And run the `vendor/bin/phpunit` command.

Some outputs are as follows:

```sh
$ php -v
PHP 8.2.0 (cli) (built: Dec 10 2022 10:52:23) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.0, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.0, Copyright (c), by Zend Technologies
$ git clone https://github.com/php-vcr/php-vcr
Cloning into 'php-vcr'...
remote: Enumerating objects: 5803, done.
remote: Counting objects: 100% (682/682), done.
remote: Compressing objects: 100% (281/281), done.
remote: Total 5803 (delta 377), reused 619 (delta 347), pack-reused 5121
Receiving objects: 100% (5803/5803), 1.10 MiB | 6.29 MiB/s, done.
Resolving deltas: 100% (3235/3235), done.
$ cd php-vcr/
$ php ~/composer.phar update
Loading composer repositories with package information
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies
Lock file operations: 71 installs, 0 updates, 0 removals
  - Locking beberlei/assert (v3.3.2)
........
$ vendor/bin/phpunit
Segmentation fault (core dumped)
```

higidi:
Ok, i'll check, maybe it is already fixed with v1.6.2 released a few minutes ago.

higidi:
Ok, i can confirm it crashes with an `Segmentation fault`. :-(
I'll spent some time in the next days. Maybe i find something or it is a bug in PHP 8.2. Seems to be the first release.

peter279k:
I think I confirm it's the internal bug error in the `PHP 8.2` version, and my debugging trace steps with `gdb` are as follows:

```sh
$ gdb php
(gdb) run vendor/bin/phpunit
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /usr/bin/php vendor/bin/phpunit
[Thread debugging using libthread_db enabled]
[...]

PR #387, opened 2023-02-20
TASK: Drop support for php 8.2

higidi:
### Context

Currently, the php-vcr library does not work well with php 8.2 so we drop the support for now until it is fixed.

### What has been done

- drop support for php >= 8.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


ecourtial:
What it the issue with PHP 8.2?

higidi:
See #373

Issue #398, opened 2023-11-27
PHP 8.3 Support


Justintime50:
Can we get PHP 8.3 support please? Currently the only thing blocking that to my knowledge is the lock of not allowing PHP 8.3 in `composer.json` I can confirm that replaying and recording cassettes seems to work on the last version of PHP VCR that allowed 8.3.

ecourtial:
https://github.com/php-vcr/php-vcr/pull/399

PR #399, opened 2023-12-01
WIP - (tech) add PHP 8.3 support

ecourtial:
### What has been done

-
-

### How to test

-
-

### Todo

- [ ]
- [ ]

### Notes

-
-


SharkMachine:
Can we get this merged? It has just been collecting dust for 3 weeks now.

composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": "^8,<8.2|>=8.2.9,<8.3",
+        "php": "^8,<8.2|>=8.2.9,<8.4",
```
Justintime50, line 19: thought: it may be better to not place this upper bound on the minor version, maybe we do the major version instead? We keep running into being blocked everytime a new PHP version comes out, may be better if people self regulated the higher end minor to minor.

```suggestion
        "php": "^8,<8.2|>=8.2.9,<9.0",
```

higidi:
fixes #398

composer.json
SharkMachine: I think it would be better to just exclude unsupported versions unless you're going to react faster to requests like this.

Justintime50: Yeah I was heavily involved in the linked thread. My main concern is by opening up the upper bound, the community can better troubleshoot and react to future PHP releases instead of needing to wait the long periods of time we have for upstream changes.

PR #401, opened 2024-02-07
FIX: Add handling for array values in response headers

SlimDeluxe:
### Context
The problem exists with servers that return duplicate headers. It's true that they should fix their own stuff, but it's also of interest that this package is resilient to such edge cases that appear in production.

### What has been done
- Fixed `Response.php` so that array as a value for any header is handled without errors.
- Made sure `Response::getContentType` still works if Content-Type is received as an array. It returns the first value of the array.
- Added 2 tests.

### How to test
`composer test`

### DIFF
--- a/src/VCR/Response.php
+++ b/src/VCR/Response.php
@@ -84,6 +84,7 @@ class Response
         $body = $response['body'] ?? null;
 
         $gzip = isset($response['headers']['Content-Type'])
+            && \is_string($response['headers']['Content-Type'])
             && str_contains($response['headers']['Content-Type'], 'application/x-gzip');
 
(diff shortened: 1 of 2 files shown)


higidi:
Thanks for the fix and for surfacing this — duplicate response headers are a real production issue and the array-preserving approach here is the right direction (only RFC-safe option for `Set-Cookie`, since duplicates must never be comma-joined).

Closing in favor of #478, which builds on this idea but keeps `getHeader()`'s existing `?string` return type (this PR's signature change would have been a breaking change) and also covers the stream_wrapper replay path, where a duplicated header was still rendered as the literal string `Array`. Full context in #477.

PR #403, opened 2024-02-23
1.5.5 parse response undefined offset 1

bwood:
### Context
This is a re-roll of https://github.com/php-vcr/php-vcr/pull/302 to achieve a patch that is compatible with php-vcr 1.5.5.

```      
```

### What has been done

### How to test

### Notes


higidi:
Closing as stalled — duplicate of PR #302.

This is a re-roll of #302 targeting an older 1.5.5 release. PR #302 is the canonical fix for `HttpUtil::parseResponse` against master. Please follow up there or in issue #272.

PR #404, opened 2024-02-23
Re-roll pull/302 patch for compatibility with php-vcr 1.5.5.

bwood:
### Context
This is a re-roll of https://github.com/php-vcr/php-vcr/pull/302 to achieve a patch that is compatible with php-vcr 1.5.5.

```      
```

### What has been done

### How to test

### Notes


higidi:
Closing as stalled — duplicate of PR #302.

This targets the legacy 1.5 branch (milestone `1.5` is closed). PR #302 is the canonical fix against master. The 1.5 branch is no longer maintained — please use a current release.

PR #410, opened 2024-02-28
Remove trailing semicolon in Content-Type header

alexzabolotny:
### Context
One of the SOAP servers we use throws an error when it sees a trailing semicolon in the header value.

### What has been done

- Remove a trailing semicolon from "Content-Type" header in SoapHook. 

### How to test

- Unfortunately, the WSDL and the whole API is behind a firewall, so it's impossible to provide the working test case.
- Fortunately, the change is not against the HTTP specification.


azabolotnyi:
@higidi @adri any chance this can be merged before summer holidays? 🥺

alexzabolotny:
@higidi @adri any chance this can be merged before summer holidays? 🥺

Issue #329, opened 2020-11-15
Symfony HttpClient compatibility


pvgnd:
php-vcr does not seems to work with Symfony HttpClient. Here is the error:  
`RuntimeException: Unexpected error, could not find curl_getinfo in response or errors`

pvgnd:
I added integration tests but any help would be appreciated to identify the issue preventing php-vcr to work with Symfony HttpClient

nicodemuz:
Seems like this call in Symfony client is causing the error for me:
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/HttpClient/Response/CommonResponseTrait.php#L143

StephanMeijer:
On my end no single request is being captured using Symfony's HTTP Client

pelletiermaxime:
Any chance that PRs #285 #286 #294 could be merged to support Symfony's http client ?

pelletiermaxime:
Hmmm well, I quickly tried the patches and they didn't fix symfony's client for me. I'm just going to use guzzle as my client for my library tests for now as I don't have time to debug this :/

cluster28:
Hi everyone,

My problem is during the multi performing in this [line](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpClient/Response/CurlResponse.php#L229) because there is not any response yet and I get this [Exception](https://github.com/php-vcr/php-vcr/blob/master/src/VCR/LibraryHooks/CurlHook.php#L276)

Hope we can fix that.

Thank you

ahmed-bhs:
@adri do you have any hints on how we might fix this ? If so, maybe we could contribute a solution.
Thanks in advance!

higidi:
PR #443 adds functional integration tests for `CurlHttpClient` and `NativeHttpClient`. Both sets of record/replay tests are currently skipped because they reproduce the bugs this issue describes:

**CurlHttpClient:** The `curl_getinfo()` call happens before `curl_multi_exec()` completes — reproduces the `RuntimeException: Unexpected error, could not find curl_getinfo in response or errors` from the issue description (CurlHook.php:280).

**NativeHttpClient:** `NativeHttpClient` passes HTTP headers as an array in the PHP stream context. `StreamHelper::createRequestFromStreamContext()` forwards this to `HttpUtil::parseRawHeader()` which expects a string, resulting in a `TypeError`. This is likely a second manifestation of the same compatibility gap.

[...]

PR #433, opened 2026-06-07
fix(http): emit \r\n line terminators in cURL header strings

higidi:
## Summary
- Header lines and status line emitted by `HttpUtil::formatHeadersForCurl`
  and `HttpUtil::formatAsStatusString` now end with `\r\n`, matching real
  cURL behaviour.
- `CurlHelper::handleOutput` reuses the same terminator for the
  end-of-headers sentinel passed to `CURLOPT_HEADERFUNCTION`.

## Background
Refs #329. Parsers such as `symfony/http-client::Response\CurlResponse`
locate the response body by the `\r\n` between header and body sections;
php-vcr emitted bare lines, breaking the replay. This adopts PR #294 by
@clemherreman (open since 2020-01).

## Notes
- The original commit (6dd0af7) is preserved with @clemherreman as
  author; cherry-pick conflicts (array-branch in `formatHeadersForCurl`,
  changed test paths) resolved in-place.
- A follow-up commit aligns the remaining unit-test expectations.
- Technically a behaviour change in `HttpUtil::formatHeadersForCurl`
  and `formatAsStatusString`, but any consumer relying on the prior
  bare-line output would already be incompatible with real cURL — we
  treat this as a bug fix rather than a BC break, suitable for the 1.9
  milestone.

## Test plan
- [x] Unit suite green: `docker compose run --rm workspace80 composer test`
- [x] phpstan/cs/ec green
- [x] Matrix sanity on PHP 8.1–8.4

### DIFF
--- a/src/VCR/Util/CurlHelper.php
+++ b/src/VCR/Util/CurlHelper.php
@@ -54,7 +54,7 @@ class CurlHelper
         if (isset($curlOptions[\CURLOPT_HEADERFUNCTION])) {
             $headerList = [HttpUtil::formatAsStatusString($response)];
             $headerList = array_merge($headerList, HttpUtil::formatHeadersForCurl($response->getHeaders()));
-            $headerList[] = '';
+            $headerList[] = "\r\n";
             foreach ($headerList as $header) {
                 self::callFunction($curlOptions[\CURLOPT_HEADERFUNCTION], $ch, $header);
             }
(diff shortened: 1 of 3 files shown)


clemherreman:
Hi! This is insane, a 7 year old PR merged :open_mouth: 

Kudos to you @higidi for bringing that project back to life, and even fixing frigging 8yo bugs. This is definitely not something I ever thought :d

PR #413, opened 2024-07-15
Various fixes from previous PRs

specialtactics:
### Context
These are our previous fixes from some previous PRs, I think they were overriding the composer.json package name so were not merged.

### What has been done
Fixes various issues which remain in the master branch.

### Notes

We've been maintaining a fork of this package with these fixes (at covergenius/php-vcr), however it's great to see that this package is once again maintained, we are just contributing back what has not yet been addressed since our fork.

PR #467, opened 2026-07-12
fix: parse responses after proxy acknowledgements

### LINKED ISSUE #456: HttpUtil::parseResponse fails on responses with multiple HTTP/ status prefixes (proxy ack scenario)
## php-vcr version

master (currently unreleased, post-1.7.x)

## PHP version

All supported (8.0–8.5)

## Library hook in use

curl

## Storage backend

Not sure (parsing happens before storage)

## Description

Responses routed through HTTP proxies can include multiple `HTTP/...` status lines before the final response — for example a `200 Connection established` tunnel acknowledgement from the proxy followed by the actual `201 Created`. The current implementation in `HttpUtil::parseResponse()` only strips the literal sequence `HTTP/1.1 100 Continue\r\n\r\n`; any other intermediate `HTTP/` prefix survives and ends up parsed as body content, corrupting both the status line and the body.

## How to reproduce

```php
use VCR\Util\HttpUtil;

// Proxy ACK (200) followed by a real response (201)
$raw = "HTTP/1.1 200 OK\r\n\r\nHTTP/1.1 200 OK\r\n\r\nHTTP/1.1 201 Created\r\nContent-Type: text/html\r\nDate: Fri, 19 Jun 2015 16:05:18 GMT\r\nVary: Accept-Encoding\r\nContent-Length: 0\r\n\r\n";

[$status, $headers, $body] = HttpUtil::parseResponse($raw);

// Expected: 'HTTP/1.1 201 Created', body: ''
// Actual:   status is 'HTTP/1.1 200 OK', body contains the remaining HTTP parts
$this->assertEquals('HTTP/1.1 201 Created', $status);
$this->assertEquals('', $body);
```

A unit test version of this reproducer is in `tests/Unit/Util/HttpUtilTest.php` — adding a case for the proxy-200 prefix (in addition to the existing 100-Continue cases) will fail on current master.

## Possible Solution

Iteratively strip leading `HTTP/...\r\n\r\n` blocks until the next block no longer starts with `HTTP/`. The existing 100-Continue stripping can be generalised: split on `\r\n\r\n`, advance the pointer past any segment that starts with `HTTP/`, then treat the remainder as header + body. Avoid split/rejoin tricks that corrupt CRLF sequences inside the body — use `substr` from the offset of the first non-HTTP block instead.

## Additional Context

This was first reported in the context of proxy tunnelling. Two earlier PRs shipped fixes but neither landed:

- PR #362 (2022) — closed 2026-06-09 as "superseded by #413"
- PR #413 (2021) — being closed now; rebase across five years of drift is more effort than reimplementing, and the body-rejoin in that patch had a separate bug (`implode('\r\n\r\n', $bodyParts)` uses the literal string, not the CRLF escape sequence)

The test cases from both PRs are preserved here as the acceptance criteria for a fresh implementation.

oiahoon:
| Q                | A
| ---------------- | ---
| Type             | Bug
| Fixes            | Fix #456
| BC break?        | no
| Deprecation?     | no
| New dependency?  | no
| License          | MIT

Parse and skip complete leading HTTP status blocks before splitting the final response. This handles proxy tunnel acknowledgements and repeated `100 Continue` blocks while preserving the final body bytes. A valid status-line boundary prevents ordinary HTTP-looking body content from being consumed.

Local validation:

- Focused `HttpUtilTest`: 12 tests, 26 assertions
- Full unit suite: 357 tests, 549 assertions
- Full integration suite: 58 tests, 331 assertions
- PHPStan: no errors
- PHP-CS-Fixer: 100 files, no fixable files
- EditorConfig: 116 files, no violations
- `composer validate --strict`, PHP syntax checks, and `git diff --check`

The Docker matrix was not run locally because Docker is unavailable in this environment. The repository CI matrix remains the source of truth for PHP 8.0–8.5 and lowest/highest dependency coverage.

### DIFF
--- a/src/VCR/Util/HttpUtil.php
+++ b/src/VCR/Util/HttpUtil.php
@@ -77,9 +77,33 @@ class HttpUtil
      */
     public static function parseResponse(string $response): array
     {
-        $response = str_replace("HTTP/1.1 100 Continue\r\n\r\n", '', $response);
+        $responseOffset = 0;
+        $separatorOffset = strpos($response, "\r\n\r\n");
+        while (false !== $separatorOffset) {
+            $rawHeader = substr($response, $responseOffset, $separatorOffset - $responseOffset);
+            $headerLines = explode("\r\n", $rawHeader);
+            $statusLine = array_shift($headerLines);
+            if (1 !== preg_match('/^HTTP\/\S+ (\d{3})(?: [^\r\n]*)?$/D', $statusLine, $matches)) {
+                break;
+            }
 
-        [$rawHeader, $rawBody] = explode("\r\n\r\n", $response, 2);
+            $statusCode = (int) $matches[1];
+            $isInformational = 100 <= $statusCode && 200 > $statusCode;
+            $isProxyAcknowledgement = 200 <= $statusCode && 300 > $statusCode && [] === $headerLines;
+            if (!$isInformational && !$isProxyAcknowledgement) {
+                break;
+            }
+
+            $nextResponseOffset = $separatorOffset + 4;
+            if ($nextResponseOffset !== strpos($response, 'HTTP/', $nextResponseOffset)) {
+                break;
+            }
+
+            $responseOffset = $nextResponseOffset;
+            $separatorOffset = strpos($response, "\r\n\r\n", $responseOffset);
+        }
+
+        [$rawHeader, $rawBody] = explode("\r\n\r\n", substr($response, $responseOffset), 2);
 
         // Parse headers and status.
         $headers = self::parseRawHeader($rawHeader);
--- a/tests/Unit/Util/HttpUtilTest.php
+++ b/tests/Unit/Util/HttpUtilTest.php
@@ -97,6 +97,26 @@ final class HttpUtilTest extends TestCase
         $this->assertEquals($expectedHeaders, $headers);
     }
 
+    public function testParseResponseAfterProxyAcknowledgements(): void
+    {
+        $raw = "HTTP/1.0 200 Connection established\r\n\r\nHTTP/2 200\r\n\r\nHTTP/1.1 201 Created\r\nContent-Type: text/plain\r\nContent-Length: 15\r\n\r\nfirst\r\n\r\nsecond";
+        [$status, $headers, $body] = HttpUtil::parseResponse($raw);
+
+        $this->assertEquals('HTTP/1.1 201 Created', $status);
+        $this->assertEquals(['Content-Type: text/plain', 'Content-Length: 15'], $headers);
+        $this->assertEquals("first\r\n\r\nsecond", $body);
+    }
+
+    public function testParseResponsePreservesBodyStartingWithHttpStatusLine(): void
+    {
+        $raw = "HTTP/1.1 200 OK\r\nContent-Type: message/http\r\nContent-Length: 35\r\n\r\nHTTP/1.1 404 Not Found\r\n\r\nreal body";
+        [$status, $headers, $body] = HttpUtil::parseResponse($raw);
+
+        $this->assertEquals('HTTP/1.1 200 OK', $status);
+        $this->assertEquals(['Content-Type: message/http', 'Content-Length: 35'], $headers);
+        $this->assertEquals("HTTP/1.1 404 Not Found\r\n\r\nreal body", $body);
+    }
+
     public function testParseHeadersBasic(): void
     {
         $inputArray = [


src/VCR/Util/HttpUtil.php
```diff
@@ -77,9 +77,25 @@ public static function parseStatus(string $status): array
...
+        $separatorOffset = strpos($response, "\r\n\r\n");
+        while (false !== $separatorOffset) {
+            $nextResponseOffset = $separatorOffset + 4;
+            $nextSeparatorOffset = strpos($response, "\r\n\r\n", $nextResponseOffset);
+            if (false === $nextSeparatorOffset) {
+                break;
+            }
+
+            $nextHeader = substr($response, $nextResponseOffset, $nextSeparatorOffset - $nextResponseOffset);
+            if (1 !== preg_match('/^HTTP\/\S+ \d{3}(?: [^\r\n]*)?(?:\r\n|$)/', $nextHeader)) {
```
higidi: This discriminator can't tell a proxy/interim preamble from a real response whose body simply begins with a valid status line. For input

```
HTTP/1.1 200 OK\r\nContent-Type: message/http\r\n\r\nHTTP/1.1 404 Not Found\r\n\r\nreal body
```

the loop advances past the real status + headers and returns status `HTTP/1.1 404 Not Found`, headers `[]`, body `real body` — the real `200` and `Content-Type` are dropped. The previous literal `str_replace` never touched the body, so this is a silent regression. Consider only skipping status-line-only preamble blocks (1xx / CONNECT 2xx with no entity) and stopping at the first block that carries headers/body.

tests/Unit/Util/HttpUtilTest.php
```diff
@@ -97,6 +97,26 @@ public function testParseContinuePlusResponseMultipleHeaders(): void
...
+        [$status, $headers, $body] = HttpUtil::parseResponse($raw);
+
+        $this->assertEquals('HTTP/1.1 201 Created', $status);
+        $this->assertEquals(['Content-Type: text/plain', 'Content-Length: 15'], $headers);
+        $this->assertEquals("first\r\n\r\nsecond", $body);
+    }
+
+    public function testParseResponsePreservesBodyStartingWithHttpText(): void
+    {
+        $raw = "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nContent-Length: 24\r\n\r\nHTTP/ is body\r\n\r\ncontent";
```
higidi: This body (`HTTP/ is body`) fails the `\S+ \d{3}` part of the regex (a space right after `HTTP/`), so the test passes trivially and never exercises the ambiguous path. Please add a case where the body starts with a valid status line, e.g. `HTTP/1.1 404 Not Found\r\n\r\ncontent`, and assert the outer `200` status and headers are preserved — that case currently fails.

higidi (changes requested):
Thanks for picking this up, @oiahoon — the `substr`-based offset approach is the right direction and correctly avoids the CRLF body-corruption bug from the earlier attempt (#413).

Requesting changes on one correctness regression plus the test meant to guard it:

1. **A response whose body starts with a valid status line is misparsed.** The loop skips a leading block whenever the *following* block looks like a status line, so a legitimate response whose body begins with `HTTP/x NNN ...` and contains a later `\r\n\r\n` loses its real status and headers. The old literal `str_replace` never inspected the body, so this is a regression — and it fails silently, getting recorded permanently into the cassette.
2. **The regression test doesn't cover this.** `testParseResponsePreservesBodyStartingWithHttpText` uses `HTTP/ is body`, which fails the `\S+ \d{3}` part and passes trivially.

Suggested direction: only treat a leading block as skippable when it is itself an interim/preamble response (a status-line-only block — a 1xx, or a CONNECT `2xx` with no entity), and stop at the first block that carries entity headers/body, rather than deciding based on what the *next* block looks like. That keeps the guarantee that response bodies are never scanned for status-line-shaped content. Please also add the test case above.

See inline comments for the exact reproducer and locations.

src/VCR/Util/HttpUtil.php
oiahoon: Addressed in 4cd8785. The loop now classifies the current block and only skips informational or headerless 2xx preambles; a response carrying headers is preserved with its body.

higidi: Confirmed. The reworked loop now classifies the current leading block and only skips 1xx informational or headerless 2xx (CONNECT) preambles, breaking as soon as a block carries entity headers. The reproducer `HTTP/1.1 200 OK\r\nContent-Type: message/http\r\n\r\nHTTP/1.1 404 Not Found\r\n\r\nreal body` now returns the `200` with its `Content-Type` intact and the body preserved. Resolving.

tests/Unit/Util/HttpUtilTest.php
higidi: The added `testParseResponsePreservesBodyStartingWithHttpStatusLine` uses a valid status line (`HTTP/1.1 404 Not Found`) inside the body and asserts the outer `200` and its headers survive, so it genuinely exercises the previously ambiguous path. Resolving.

higidi (approved):
Both points from the previous round are addressed: the preamble skip is now driven by classifying each leading block (1xx / headerless 2xx CONNECT ack) rather than by the shape of the following block, so a response whose body begins with a valid status line is preserved. The regression test now covers that path, and CI is green across the full 8.0–8.5 × lowest/highest matrix. Thanks @oiahoon.

PR #494, opened 2026-08-06
Update PHPUnit guide

angelov:
| Q                | A
| ---------------- | ---
| Type             | Documentation 
| Fixes            | /
| BC break?        | no 
| Deprecation?     | no 
| New dependency?  | no 
| Docs updated?    | yes 
| License          | MIT

The `angelov/phpunit-php-vcr` package has been out for few years, and supports latest versions of PHPUnit, so I propose it's listed in the official guide.

### DIFF
--- a/docs/howto/use-with-phpunit.md
+++ b/docs/howto/use-with-phpunit.md
@@ -1,47 +1,61 @@
 # Use with PHPUnit
 
-> One-liner: call VCR manually per test — turn it on, insert a cassette, turn it off.
+> **One-liner:** use the [`angelov/phpunit-php-vcr`](https://github.com/angelov/phpunit-php-vcr) PHPUnit extension to automatically manage PHP-VCR and cassettes with the `#[UseCassette]` attribute.
 
-**On this page:** [Manual lifecycle](#manual-lifecycle)
+**On this page:** [Installation](#installation) · [Recording requests](#recording-requests)
 
-## Manual lifecycle
+## Installation
+
+Install the PHPUnit integration:
+
+```bash
+composer require --dev angelov/phpunit-php-vcr
+```
+
+Then register the extension in your `phpunit.xml`:
+
+```xml
+<extensions>
+    <bootstrap class="\Angelov\PHPUnitPHPVcr\Extension">
+        <parameter name="cassettesPath" value="tests/fixtures" />
+    </bootstrap>
+</extensions>
+```
+
+The extension automatically configures and manages PHP-VCR for your tests. Additional configuration options (record modes, storage, request matchers, library hooks, etc.) are available in the project's documentation.
+
+## Recording requests
+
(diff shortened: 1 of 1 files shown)


higidi (changes requested):
Thanks for the PR — and for keeping the extension maintained across PHPUnit 10 through 13. It belongs in this guide.

One thing I'd like to change about the shape: I'd rather **keep both paths on the page** instead of replacing the manual lifecycle. php-vcr supports PHP 8.0–8.5 and is tested against PHPUnit 9.5+, while `angelov/phpunit-php-vcr` v1.3 requires `php: ^8.2` and `php-vcr/php-vcr: ^1.6`, and the `<extensions><bootstrap>` element only exists from PHPUnit 10 on. As written, users on PHP 8.0/8.1 or PHPUnit 9 lose their only documented recipe.

Concretely:

- Keep `## Manual lifecycle` as the baseline, including `testThrowsWithoutACassette` — it documents the fail-loud behaviour when no cassette is inserted.
- Add the extension as a second section, e.g. `## With the PHPUnit extension`, with a one-line requirements note (PHP 8.2+, PHPUnit 10+).
- Keep the note about `php-vcr/phpunit-testlistener-vcr` — it explains why the old `@vcr` annotation no longer works, which readers still run into. Your new note can sit alongside it.
- Instead of "additional configuration options are available in the project's documentation", name them: `cassettesPath`, `storage`, `libraryHooks`, `requestMatchers`, `whitelistedPaths`, `blacklistedPaths`, `mode`. All optional, and it saves the reader a round-trip.

Two small things:

- The one-liner: every other page uses plain `> One-liner: …` without bold — please match that.
- The bare URL at the end of the note trips markdownlint `MD034/no-bare-urls`, which fails the Documentation CI job (reproduced locally with this repo's config). An inline link fixes it and matches the rest of `docs/`, which has no bare URLs today:

```markdown
> **📌 Note:** … see the [project README](https://github.com/angelov/phpunit-php-vcr) for the full documentation.
```

Finally, a bigger question, and the reason I deliberately left the old test-listener package unmaintained: I've been considering **bringing the PHPUnit integration into php-vcr itself**. Code-wise it isn't much, and having it in one place makes it far easier to maintain and to keep in sync with the library's own supported PHP/PHPUnit range. You've been carrying that integration for years — what do you think of that direction, and would you be interested in doing it here? Happy to open a dedicated issue and put it in a milestone if you're up for it. Either way this docs change can land on its own; it's not a precondition.

angelov:
Thank you for the feedback @higidi! I've updated the file to keep the existing flow explanation.

As for the other question, I see your point and I agree. The package itself doesn't really require much maintenance, so I am okay with any future direction - integrating it in this repo, or keeping it as a separate package (either as it is nowadays, or moved to the php-vcr org). Considering that other people have contributed to the extension, I wonder if we should do any specific steps or something regarding the license, if we choose to integrate it here?

Issue #446, opened 2026-06-09
Refactor request matchers to objects with cassette-level scope (RequestMatcherInterface)


higidi:
## Idea

Today request matchers are plain PHP callbacks registered globally in `Configuration`. Two improvements have been proposed in PR #266:

1. **`RequestMatcherInterface`** — make matchers typed objects instead of raw callbacks, improving type safety and IDE support.
2. **Cassette-scoped matchers** — store matchers per cassette instead of globally, so different cassettes can use different matching strategies without global state changes.

## Motivation

Complex matching scenarios (e.g. "match all requests to `/product` except when `id=42`, which should match on query string too") currently require a single global custom callback that encodes all the logic. Cassette-scoped matchers would allow per-request matcher overrides directly in the cassette definition.

## BC break

Both changes are backwards-incompatible:
- `RequestMatcherInterface` replaces the current callable type
- Cassette-stored matchers change the cassette format

Target: 2.0 only.

## Reference

See PR #266 for a prior implementation attempt. The PR is too old to merge directly — the codebase has changed significantly — but the concept is sound.

PR #505, opened 2026-08-15
feat(matcher): add body_json request matcher for semantic JSON body comparison

### LINKED ISSUE #506: Semantic JSON body matching: reordered object keys break replay
### Description

`RequestMatcher::matchBody()` compares request bodies as raw strings, so for JSON APIs replay depends on how
a payload was *serialised* rather than on what it means. Two requests a server cannot tell apart fail to
match whenever:

- an object key order changes — a refactor moving a field, a serialiser that sorts keys, an SDK upgrade;
- the payload is pretty-printed on one side and minified on the other;
- whitespace differs anywhere in the document.

The failure is unhelpful in both record modes. With `MODE_NONE` the test dies with the "does not match a
previously recorded request" `LogicException`. With `MODE_NEW_EPISODES` a real HTTP call goes out silently
and a near-duplicate recording is appended to the cassette — the cassette grows a second entry for what is
the same call.

The documented way out is to drop `body` from `enableRequestMatchers()`, but that removes body matching
altogether: the cassette then replays for *any* body, including one that genuinely differs. There is no
setting between "byte-identical" and "ignored".

This isn't specific to one API. It applies to any JSON-over-HTTP client whose serialisation order isn't
pinned — REST, GraphQL, JSON-RPC.

A JSON-aware body matcher is easy to get subtly wrong in userland, which is an argument for it living next
to the other built-ins. The obvious one-liner —
`json_decode($a, true) == json_decode($b, true)` — is wrong four ways at once: `==` is loose, so `1` matches
`"1"` and `0` matches `null`; associative decoding makes `{"0":"a"}` indistinguishable from `["a"]`; array
element order silently stops mattering; and invalid or empty bodies need a defined fallback.

### Example

```php
\VCR\VCR::configure()
    ->enableRequestMatchers(['method', 'url', 'host', 'query_string', 'body_json']);
```

Before — recorded body and incoming body are the same document, replay fails:

```php
// recorded: {"model":"a","stream":false}
// incoming: {"stream":false,"model":"a"}
// -> no match, LogicException (MODE_NONE) or a live call + duplicate recording (MODE_NEW_EPISODES)
```

After, with a matcher that compares the body as a JSON document:

```php
// {"model":"a","stream":false}  vs.  {"stream":false,"model":"a"}  -> match
// ["a","b"]                     vs.  ["b","a"]                     -> no match  (array order is significant)
// {"n":1}                       vs.  {"n":"1"}                     -> no match  (scalars stay strictly typed)
// {}                            vs.  []                            -> no match  (object is not array)
```

Bodies that are empty, invalid JSON, or a bare scalar would fall back to the existing raw-string comparison,
so nothing that matches today stops matching.

Two properties seem worth stating up front, since they're what makes this useful rather than just lenient:
array ordering must stay significant (reordering a list changes the request), and scalar types must stay
strict. A matcher that ignores both is too blunt to trust.

### Documentation impact

- `docs/reference/request-matchers.md` — new matcher entry, the "All 8" count, the on-this-page nav.
- `docs/guides/request-matching.md` — the count, and a pointer in *Narrowing what matters* noting this as
  the targeted alternative to dropping `body` when the variation is only in JSON serialisation.

---

I have an implementation ready and opened it as a draft PR (#505) before finding the right place to start
the discussion — happy for the design conversation to happen here and to reshape or close the PR based on
it. Two points in it are worth your opinion specifically:

1. Registering the matcher in `$availableRequestMatchers` means it joins the implicit "all enabled" default
   set. The matching *outcome* is provably unchanged (it's ANDed with `body`, which already rejects anything
   it would reject), and there's a regression test asserting exactly that — but if you'd rather the default
   set stay literally untouched, it can be held in a separate list consulted only by name resolution.
2. It's deliberately scoped to a boolean matcher. Reporting *where* two JSON bodies diverge is a separate
   concern and I haven't mixed it in.

I've also kept it on the current `callable(Request, Request): bool` architecture rather than anticipating
#446, so whatever migration you apply to the eight existing matchers applies to this one unchanged.

mohaelmrabet:
| Q                | A
| ---------------- | ---
| Type             | Feature
| Fixes            | Fix #506
| BC break?        | no
| Deprecation?     | no
| New dependency?  | no
| Docs updated?    | yes
| License          | MIT

Adds a `body_json` request matcher that compares request bodies as JSON documents instead of raw strings.

`matchBody()` compares bytes, so replay depends on the *serialisation* of a JSON payload rather than on its
meaning. A reordered object key, a reformatted document or a switch between pretty-printed and minified JSON
all break replay for a request the server cannot tell apart. The failure is unhelpful in both record modes:
with `MODE_NONE` the test dies, with `MODE_NEW_EPISODES` a live HTTP call goes out and a near-duplicate
recording is appended. The usual workaround — dropping `body` from `enableRequestMatchers()` — throws away
body matching entirely, which is exactly the discrimination the user wanted.

```php
\VCR\VCR::configure()
    ->enableRequestMatchers(['method', 'url', 'host', 'query_string', 'body_json']);
```

```php
// {"model":"a","stream":false}  vs.  {"stream":false,"model":"a"}  -> match
// ["a","b"]                     vs.  ["b","a"]                     -> no match
// {"n":1}                       vs.  {"n":"1"}                     -> no match
// {}                            vs.  []                            -> no match
```

Object key order is ignored, array element order stays significant, scalars are compared strictly, and key
sets must be equal (it is not a subset matcher). When either body is empty, invalid JSON or a bare scalar it
falls back to the existing `matchBody()` string comparison.

### Implementation notes

Two decisions are worth flagging because the obvious alternatives are subtly wrong:

- Bodies are decoded to `stdClass` rather than in associative mode. `json_decode('{"0":"a"}', true)` and
  `json_decode('["a"]', true)` produce the very same PHP value, which would make an object
  indistinguishable from an array. `array_is_list()` would not have helped, and it needs PHP 8.1 anyway.
- `JSON_BIGINT_AS_STRING` is applied to both sides, so two different integers beyond `PHP_INT_MAX` cannot
  collapse to the same float and match by accident.

Both are covered by tests (`numeric keys object against array`, `big integers beyond php int max`).

### Does this change the default matcher set?

`body_json` is registered in `Configuration::$availableRequestMatchers`, so with the default configuration
(`enabledRequestMatchers === null`) it joins the enabled set. **The matching outcome is unchanged**, because
matchers are ANDed and `body_json` is strictly weaker than `body`:

- if `body` returns `true` the bodies are the identical string, therefore semantically equal, therefore
  `body_json` returns `true`;
- if `body` returns `false` the AND is already `false`.

So `body AND body_json ≡ body`. `testDefaultMatcherSetOutcomeIsUnchangedByBodyJson` asserts this rather than
leaving it as a claim: it runs the full default set against a table of body pairs and compares the result
with and without `body_json` enabled. The entry is also listed **last** in the array, so `body` short-circuits
before anything gets decoded on the failing path.

If you would rather the implicit "all" default stay literally untouched, the matcher can be held in a
separate list consulted only by `enableRequestMatchers()`/`addRequestMatcher()` name resolution — a ~10-line
`Configuration` change that leaves `RequestMatcher` as-is. Happy to switch if you prefer that.

### Backward compatibility

- Public API unchanged; one static method added to `VCR\RequestMatcher`.
- Default behaviour unchanged (asserted, see above).
- Cassette format untouched — this PR only affects matching. Cassettes recorded by any earlier version
  replay identically.
- No new dependency. `ext-json` is always available on PHP 8+ and `Storage\Json` already relies on it.
- PHP 8.0 compatible: only `json_decode`, `json_last_error`, `get_object_vars` and `strspn` are used.

### Docs

- `docs/reference/request-matchers.md` — new `body_json` entry with a `🆕 Since 1.13` callout, nav line and
  the "All 8" → "All 9" count.
- `docs/guides/request-matching.md` — the count, plus a pointer in *Narrowing what matters* explaining that
  swapping `body` for `body_json` is the targeted alternative to dropping body matching.

Both documented snippets are exercised by tests: the `enableRequestMatchers([...])` example and the
key-reordering behaviour are what `CassetteTest::testPlaybackWithBodyJsonMatcherIgnoresObjectKeyOrder`
records and replays through a real `Cassette` + `Yaml` storage.

### Test plan — what I actually ran

40 new cases: 31 in `RequestMatcherTest` across two data providers (reordered object keys, reformatting,
nesting, array order at depth, strict scalar typing, object-vs-array, numeric-key objects, extra/missing
keys, array length, invalid JSON, one-sided JSON, absent and empty bodies, whitespace-only and
leading-whitespace bodies, big integers, bare scalar bodies, XML and binary bodies), 1 in
`ConfigurationTest` for registration plus the 7-case default-set regression provider, and 1 in
`CassetteTest` covering record-then-replay end to end.

Docker matrix, run sequentially, each after its own `composer update`:

| Workspace | Deps | Result |
| --- | --- | --- |
| workspace80 | highest | 595 tests, OK |
| workspace80 | lowest | 595 tests, OK |
| workspace85 | highest | 595 tests, OK |
| workspace85 | lowest | 595 tests, OK |

`composer phpstan`, `composer cs` and `composer ec` on workspace80: all exit 0, no new baseline entries,
`cs-fix` reported 0 of 127 files to fix.

The matrix was re-run in full after the `perf` commit; the numbers above are from that run.

One thing to call out honestly: on workspace85 the run reports 31 PHPUnit deprecations against 28 on an
unmodified `master`. The 3 extra ones are my 3 new `@dataProvider` docblock annotations — I followed the
convention already used throughout `tests/`, so they will disappear with the suite-wide move to attributes
rather than needing anything special here. Say the word if you would prefer attributes in this PR.

Opened #506 with the problem statement and the design questions — happy for the discussion to happen there
rather than in this diff.

### DIFF
--- a/docs/guides/request-matching.md
+++ b/docs/guides/request-matching.md
@@ -7,8 +7,8 @@
 
 ## The default: everything must agree
 
-By default, all 8 built-in matchers are enabled — method, URL path, host, headers, body, post fields, query
-string, SOAP operation. An incoming request only matches a recording if **every enabled matcher** returns
+By default, all 9 built-in matchers are enabled — method, URL path, host, headers, body, JSON body, post
+fields, query string, SOAP operation. An incoming request only matches a recording if **every enabled matcher** returns
 `true` (logical AND, see [`Request::matches()`](../reference/request-response.md#request)). That's the
 strictest possible interpretation of "same request", and it's the safest default: nothing replays unless it's
 genuinely the same call.
@@ -25,6 +25,16 @@ identify the request for your use case:
 
 Now two requests that only differ in headers or body replay the same recording. See the full list and exact
 semantics of each in the [Request Matchers reference](../reference/request-matchers.md).
+
+Dropping `body` is a blunt instrument when the variation is only in how a JSON payload is serialised — a
+reordered key, a reformatted document. Swap it for
+[`body_json`](../reference/request-matchers.md#body_json) instead, which compares the body as a JSON
+document rather than as a string, and keeps rejecting bodies that actually differ:
+
+```php
+\VCR\VCR::configure()
+    ->enableRequestMatchers(['method', 'url', 'host', 'query_string', 'body_json']);
+```
 
 ## Custom matchers
 
--- a/docs/reference/request-matchers.md
+++ b/docs/reference/request-matchers.md
@@ -3,9 +3,9 @@
 > One-liner: matching decides whether an incoming request "is" a recorded one. All enabled matchers must
 > agree (AND) — see [Request Matching](../guides/request-matching.md) for the concept.
 
-**On this page:** [method](#method) · [url](#url) · [host](#host) · [headers](#headers) · [body](#body) · [post_fields](#post_fields) · [query_string](#query_string) · [soap_operation](#soap_operation) · [Custom matchers](#custom-matchers)
+**On this page:** [method](#method) · [url](#url) · [host](#host) · [headers](#headers) · [body](#body) · [body_json](#body_json) · [post_fields](#post_fields) · [query_string](#query_string) · [soap_operation](#soap_operation) · [Custom matchers](#custom-matchers)
 
-All 8 are enabled by default. Configure a subset via
+All 9 are enabled by default. Configure a subset via
 [`enableRequestMatchers()`](configuration.md#request-matchers).
 
 ## `method`
@@ -39,6 +39,34 @@ All 8 are enabled by default. Configure a subset via
 ## `body`
 
 - **Compares:** the raw request body string
+
+## `body_json`
+
+> **🆕 Since 1.13**
+
+- **Compares:** the request body decoded as JSON, structurally — object key order is ignored, array
+  element order is significant, and scalars are compared strictly.
+- Falls back to the same raw-string comparison as [`body`](#body) when either body is empty, invalid
+  JSON, or a bare scalar.
+
+```php
+// {"model":"a","stream":false}  vs.  {"stream":false,"model":"a"}  -> match
+// ["a","b"]                     vs.  ["b","a"]                     -> no match
+// {"n":1}                       vs.  {"n":"1"}                     -> no match
+// {}                            vs.  []                            -> no match
+```
+
+Replace `body` with `body_json` to get order-insensitive JSON matching:
+
+```php
+\VCR\VCR::configure()
+    ->enableRequestMatchers(['method', 'url', 'host', 'query_string', 'body_json']);
+```
+
+> **📌 Note**
+> Like every built-in, `body_json` is part of the default set — where it changes nothing, because
+> matchers are ANDed and `body` already rejects any body it would reject. Leaving `body` enabled
+> alongside it keeps the strict raw-string comparison in force.
 
 ## `post_fields`
 
--- a/docs/reference/request-matchers.md
+++ b/docs/reference/request-matchers.md
@@ -68,6 +68,15 @@ Replace `body` with `body_json` to get order-insensitive JSON matching:
 > matchers are ANDed and `body` already rejects any body it would reject. Leaving `body` enabled
 > alongside it keeps the strict raw-string comparison in force.
 
+> **⚠️ Warning**
+> Decoding is only reached when two bodies differ as strings *and* both start with `{` or `[`.
+> Identical bodies and non-JSON bodies (XML, binary uploads, form-encoded payloads) are settled by the
+> same string comparison `body` uses, so the default matcher set costs nothing extra. When decoding
+> does happen it is paid once per recording examined, against both bodies — on a 200 KiB payload that
+> is milliseconds per recording, so a cassette holding many large JSON recordings will notice.
+> Narrowing the enabled matchers so cheaper ones (`method`, `url`, `host`) reject non-candidates first
+> keeps that cost off the hot path.
+
 ## `post_fields`
 
 - **Compares:** `Request::getPostFields()` — parsed POST field array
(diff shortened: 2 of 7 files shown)


mohaelmrabet:
### Performance

Worth recording explicitly, since "decode JSON on every comparison" is a fair thing to be wary of in a
matcher that runs once per recording examined.

Measured on PHP 8.5, per call, against `matchBody()` as the 0.0010 ms baseline. A payload of 1000 chat-style
message objects gives the 214 KiB case:

| Case | Before | After |
| --- | --- | --- |
| identical bodies, 2.2 KiB | 0.0407 ms | 0.0011 ms |
| identical bodies, 214 KiB | 3.4179 ms | 0.0011 ms |
| non-JSON body, 214 KiB | 0.0321 ms | 0.0109 ms |
| reordered keys, 214 KiB | 3.4525 ms | 3.5197 ms |

The first two rows were the problem. In the **default** matcher set `body_json` is only ever reached after
`body` has already returned `true` — i.e. after the strings have been proven identical — so every user who
never asked for this matcher was paying to decode two identical documents, 3.4 ms of it on a 214 KiB body.
Comparing the raw strings first removes that entirely: the default set is back to the cost of `body`.

The third row is non-JSON bodies — XML, binary uploads, form-encoded payloads. Those were decoded before
being rejected, and `trim()` copied the whole body first. `strspn()` now walks the leading whitespace without
copying and the body is dismissed on its first non-whitespace byte.

The last row is unchanged and intended: two genuinely different JSON documents have to be decoded to be
compared. That cost is paid only by someone who enabled `body_json` deliberately, and it is small next to the
HTTP request it avoids.

**Known limitation, not addressed here:** `Cassette::playback()` walks every recording, so the *incoming*
body is decoded once per recording rather than once per playback. With many large JSON recordings in one
cassette that is wasted work. A static matcher callback has nowhere to cache it; the matcher objects with
cassette-level scope proposed in #446 would be the natural place. Narrowing the enabled matchers so `method`
/ `url` / `host` reject non-candidates first keeps it off the hot path in the meantime.

Docs updated with a `⚠️ Warning` callout covering the same ground. The benchmark script was throwaway and is
not part of the diff.

